PR #3, opened 2023-06-06
fix tokenizer and speedup with fp16

DongyaoZhu:


mindone/pipelines/text_classifiers/SequenceClassificationPipeline.py
```diff
@@ -21,30 +21,34 @@ def __init__(
...
         elif model_name == 'roberta_base':
             from mindnlp.models.roberta import RobertaConfig
             from mindnlp.transforms import RobertaTokenizer
             from mindnlp.models import RobertaForSequenceClassification
+            from mindnlp._legacy.amp import auto_mixed_precision
             pad_token_id = config.get('pad_token_id', 1)
             self.config = RobertaConfig(**config)
             self.tokenizer = RobertaTokenizer(tokenizer_path)
             self.backbone = RobertaForSequenceClassification(self.config)
+            self.backbone = auto_mixed_precision(self.backbone, 'O1')
```
SamitHuang, line 40: 这里amp_level最好是可设置，hard code O1 is not recommended.

DongyaoZhu: done

PR #8, opened 2023-06-12
bmm <2.0 compatibility

DongyaoZhu:


examples/stable_diffusion_v2/ldm/modules/diffusionmodules/model.py
```diff
@@ -157,6 +157,7 @@ def __init__(self, in_channels, dtype=ms.float32):
         super().__init__()
         self.in_channels = in_channels
         self.dtype = dtype
+        self.bmm = P.BatchMatMul()
```
SamitHuang, line 160: is it tested on ms2.0?

DongyaoZhu: ops.bmm calls P.BatchMatMul() internally, see https://www.mindspore.cn/docs/zh-CN/r2.0/_modules/mindspore/ops/function/math_func.html#bmm

PR #9, opened 2023-06-13
Fix bugs. Support SD2.0 parallel finetuning.

SamitHuang:


examples/stable_diffusion_v2/ldm/modules/encoders/text_encoder.py
DongyaoZhu, line 1: maybe change 23 to 24

SamitHuang: lead to blur/worse generation. let's fix after ckpt conversion

PR #14, opened 2023-06-15
change to paper name bert-mpu

DongyaoZhu:


SamitHuang (commented):
What is the motivation to rename it?

vigo999:
SequenceClassificationPipeline is a common name for sequence classification.
rename it to let users know about what is the actual classifier they are using

PR #33, opened 2023-07-05
Add DDIM, DPM-Solver(including++), UniPC samplers

zhaoyuzhi:


SamitHuang (approved):
looks good. further test is needed.

PR #35, opened 2023-07-05
Fix bugs (MS2.0/Ascend/Inference)

congw729:
1. Validator not exsit in ms2.0
2. Tensor data type inconsist

Optimize codes:
1. change ms context mode to static variable

### DIFF
--- a/examples/stable_diffusion_v2/text_to_image.py
+++ b/examples/stable_diffusion_v2/text_to_image.py
@@ -124,7 +124,7 @@ def main(args):
 
     # set ms context
     device_id = int(os.getenv("DEVICE_ID", 0))
-    mode = 0
+    mode = ms.context.GRAPH_MODE 
     ms.context.set_context(
         mode=mode,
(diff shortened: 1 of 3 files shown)


DongyaoZhu (approved):

DongyaoZhu:
LGTM

examples/stable_diffusion_v2/text_to_image.py
```diff
@@ -124,7 +124,7 @@ def main(args):
 
     # set ms context
     device_id = int(os.getenv("DEVICE_ID", 0))
-    mode = 0
+    mode = ms.context.GRAPH_MODE
```
DongyaoZhu, line 127: if this is the only place `mode` is used, then the variable may not be needed

congw729: Actually, most of the usage is using ms.context.GRAPH_MODE, that's why I changed this place to be consistent with other usages.

PR #50, opened 2023-07-13
add tools; add script for converting hgf sd2 weights to ms weights

DongyaoZhu:


congw729 (approved):
LGTM

mindone/tools/sd_torch_to_mindspore/convert.sh
```diff
@@ -0,0 +1,9 @@
+# export RANK_ID=0
+
+source=v2-1_768-ema-pruned.ckpt
+python pt2np_v2.py $(source)
```
congw729, line 4: $source

DongyaoZhu: thank you very much

PR #58, opened 2023-07-19
add resume training

Songyuanwei:
When training is interrupted, the resume training function can be used to continue training from the interrupted point.
For use this function, you can set the parameter `resume` is True or path to resume checkpoint.

### DIFF
--- a/examples/stable_diffusion_v2/ldm/modules/train/checkpoint.py
+++ b/examples/stable_diffusion_v2/ldm/modules/train/checkpoint.py
@@ -1,8 +1,11 @@
 """checkpoint manager """
+import logging
 import os
 import stat
 
 import mindspore as ms
+
+_logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


SamitHuang:
require further test

examples/stable_diffusion_v2/ldm/modules/train/checkpoint.py
```diff
@@ -45,9 +48,9 @@ def remove_ckpt_file(self, file_name):
                 os.chmod(file_name, stat.S_IWRITE)
                 os.remove(file_name)
         except OSError:
-            print(f"OSError, failed to remove the older ckpt file {file_name}.")
+            _logger.warning(f"OSError, failed to remove the older ckpt file {file_name}.")
```
SamitHuang, line 51: use _logger.error

SamitHuang (approved):

PR #66, opened 2023-07-28
Fix bugs in vanilla finetune

SamitHuang:


examples/stable_diffusion_v2/ldm/modules/train/ema.py
```diff
@@ -32,7 +33,10 @@ def __init__(self, network, ema_decay=0.9999, updates=0, trainable_only=True):
 
         self.hyper_map = C.HyperMap()
         self.map = ops.HyperMap()
-        self.assign = ops.Assign().add_prim_attr("primitive_target", "CPU")
+        if offloading:
+            self.assign = ops.Assign().add_prim_attr("primitive_target", "CPU")
```
DongyaoZhu, line 37: the last time I used offloading was in this way:
self.conv = nn.Conv2d(...)
self.conv.conv2d.add_prim_attr("primitive_target", "CPU")
Not sure if there's a difference.

DongyaoZhu (approved):
lgtm

examples/stable_diffusion_v2/ldm/modules/train/ema.py
SamitHuang: Why `self.conv.conv2d.add_prim_attr`  rather than `self.conv.add_prim_attr`

DongyaoZhu: an nn.Cell would suffice

PR #69, opened 2023-07-31
adapt Ascend 910B

CaitinZhao:


examples/stable_diffusion_v2/train_text_to_image.py
```diff
@@ -62,6 +62,7 @@ def init_env(args):
         device_id=device_id,
         max_device_memory="30GB",  # TODO: why limit?
     )
+    ms.set_context(ascend_config={"precision_mode": "allow_fp32_to_fp16"})    # Only effective on Ascend 901B
```
SamitHuang, line 65: It is the default setting for non-910B devices?

SamitHuang (commented):
Need code formatting fo CI check

examples/stable_diffusion_v2/train_text_to_image.py
CaitinZhao: other devices do not take effect.

SamitHuang (approved):

PR #89, opened 2023-08-17
Add convert scripts for weight of controlnet (except cond_stage model).

congw729:
Convert scripts for converting weights of (partial) ControlNet.


DongyaoZhu:
hold until pr#90 is settled

congw729:
> [outdated] hold until pr#90 is settled

PR#90 has been merged.

DongyaoZhu (commented):
does it function without cond?

congw729:
> does it function without cond?

After converting, we have to load the base model (to load the weights of FrozenClip model) first then load this converted model to function normally.

DongyaoZhu:
> > does it function without cond?
> 
> After converting, we have to load the base model (to load the weights of FrozenClip model) first then load this converted model to function normally.

can you point to the code where it's loading twice? I don't seem to find it in i2i.py

congw729:
> > > does it function without cond?
> > 
> > 
> > After converting, we have to load the base model (to load the weights of FrozenClip model) first then load this converted model to function normally.
> 
> can you point to the code where it's loading twice? I don't seem to find it in i2i.py

For now, this part is done manually. Should I add this part to the convert codes? It will require an additional input (path to the base weights file of ms).

DongyaoZhu:
> > > > does it function without cond?
> > > 
> > > 
> > > After converting, we have to load the base model (to load the weights of FrozenClip model) first then load this converted model to function normally.
> > 
> > 
> > can you point to the code where it's loading twice? I don't seem to find it in i2i.py
> 
> For now, this part is done manually. Should I add this part to the convert codes? It will require an additional input (path to the base weights file of ms).

Maybe add to i2i: --vae and --clip, with a same base ckpt as default.

SamitHuang:
Update accordkingy

PR #94, opened 2023-08-21
Make weight decay configurable and change default value

SamitHuang:
- Previous bug:  
     weight decay is fixed to **1e-6**  
     The reference value used in LAION training or LoRA fine-tune in diffusers is 1e-2 

- Fix in this PR:  weight decay is now configurable via arg parser.
         Now default weight decay is  **1e-2**

### DIFF
--- a/examples/stable_diffusion_v2/scripts/run_train_v2.sh
+++ b/examples/stable_diffusion_v2/scripts/run_train_v2.sh
@@ -9,9 +9,9 @@ device_id=0
 
 output_path=output/finetune_pokemon
 task_name=txt2img
-data_path=./datasets/diffusion/pokemon_blip/train
+data_path=./datasets/pokemon_blip/train
 #data_path=/home/yx/datasets/diffusion/pokemon
-epochs=20
+epochs=10
 warmup_steps=1000
 
 pretrained_model_path=models/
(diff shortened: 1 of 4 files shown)


examples/stable_diffusion_v2/scripts/run_train_v2.sh
```diff
@@ -9,9 +9,9 @@ device_id=0
 
 output_path=output/finetune_pokemon
 task_name=txt2img
-data_path=./datasets/diffusion/pokemon_blip/train
+data_path=./datasets/pokemon_blip/train
 #data_path=/home/yx/datasets/diffusion/pokemon
-epochs=20
+epochs=10
```
DongyaoZhu, line 14: is this related too?

examples/stable_diffusion_v2/scripts/run_train_v2.sh
```diff
@@ -47,4 +49,5 @@ python train_text_to_image.py \
     --epochs $epochs \
     --warmup_steps $warmup_steps \
     --image_filter_size=$image_filter_size \
+    --weight_decay=$weight_decay \
```
DongyaoZhu, line 52: config files or args?

DongyaoZhu (approved):
please see comments

PR #101, opened 2023-08-24
[feat] add karras sampler

heepengpeng:


SamitHuang:
Please complement https://github.com/mindspore-lab/mindone/blob/master/examples/stable_diffusion_v2/schedulers.md for the added schedulers

PR #104, opened 2023-08-25
Fix flash attention for ms2.1

VictorHe-1:
Motivation:
1. tiling strategy "xunfei" is not supported in ms2.1 release version, only "sparse" and "wukong" are supported;

Test plan:
Already tested in ModelArts platform.


zhanghuiyao (approved):

DongyaoZhu (approved):
lgtm; btw the conventional short name for "strategy" would be "strat" instead of "stgy"

PR #111, opened 2023-08-29
feat: add videocomposer

geniuspatrick:


examples/videocomposer/.gitignore
hadipash, line 1: .gitignore should be one per repo.

examples/videocomposer/LICENSE
SamitHuang, line 1: We used to claim apache 2.0 License. Why MIT now for this case

examples/videocomposer/assets/results/exp02_motion_transfer-S00009.gif
SamitHuang, line 1: Maybe just put the resulting gif to github user content server?

examples/videocomposer/clip/clip_modules.py
SamitHuang, line 1: consider re-use the CLIP models in stable_diffusion_v2 which already guarantees the graph mode and inference speed (~5s to generate a batch of images)

examples/videocomposer/demo_video/moon_on_water.jpg
SamitHuang, line 1: It would make our code cleaner to tell the users to download these source images or videos from original VideoComposer in readme instructions, instead of making a copy

examples/videocomposer/LICENSE
geniuspatrick: Original repo use MIT License. Need to check if the change is reasonable.

examples/videocomposer/clip/clip_modules.py
geniuspatrick: @wtomin

examples/videocomposer/demo_video/moon_on_water.jpg
geniuspatrick: That's right. . . We uploaded sample videos to the repository because the original author did so. We should discuss how to place these media files for easy access.

examples/videocomposer/assets/results/exp02_motion_transfer-S00009.gif
geniuspatrick: same as the demo video files

examples/videocomposer/.gitignore
hadipash, line 1: There should be one `.gitignore` per repo. Merge it with the main, if necessary.

geniuspatrick: In fact, it is common for each sub-repo to have its own `.gitignore` configuration for such a collection of repositories. I've added some necessary rules here to override the parent directory and keep the root configuration intact.
I would suggest minimal pruning of the root configuration in the future, e.g. removing entries such as `data`, `*.txt`, etc. and making specific ignore rules only under each subrepository.

examples/videocomposer/clip/clip_modules.py
wtomin: The current repo contains just a copy of the CLIP models codes in stable_diffusion_v2 plus a minor change on the CLIPtokenizer (to make it process empty string). It is possible to refactor the code to make it concise. I will make changes to it.

examples/videocomposer/demo_video/moon_on_water.jpg
wtomin: My suggestion is: for the video/images in the `demo_video/`, we should put it in the cloud and provide permenant download links; while for gif results and illustration images in readme files, we can keep these images in github user content server, because these images maybe replaced occasionally.

SamitHuang (approved):
Overall it's good as a basic version supporting pynative inference.  We can merge it to have a starting point.

wtomin:
Before merging it to the main branch, can we just reduce the overall size? Otherwise, downloading this repo in some server with bad connection is an endless job. @geniuspatrick

wtomin (approved):

examples/videocomposer/assets/results/exp02_motion_transfer-S00009.gif
hadipash: > Maybe just put the resulting gif to github user content server?

I agree. Cloning this many heavy assets will take some time.

PR #121, opened 2023-09-05
add eval fvd and kvd for videos

Songyuanwei:
currently, it's can only run on 910A with ms2.1.


DongyaoZhu (approved):
readme?

SamitHuang:
Put them into `mindone/` instead of `stable_diffusion_v2/.`

PR #149, opened 2023-09-19
Add graph mode support for controlnet with inference framework

Gaohan123:


examples/stable_diffusion_v2/cldm/cldm.py
```diff
@@ -60,6 +60,77 @@ def construct(self, x, timesteps=None, context=None, control=None, only_mid_cont
         return self.out(h)
 
 
+class ControlnetUnetModel(UNetModel):
+    def __init__(self, control_stage_config, guess_mode=False, strength=1.0, *args, **kwargs):
```
congw729, line 64: Why not just edit ControledUnetModel?

Gaohan123: Because it will influence original inference pipeline. They have a large difference

SamitHuang (approved):

congw729 (approved):
LGTM

PR #153, opened 2023-09-21
Fix 910b bugs (on MS master 0919); Allow grad accumulation setting; Fix checkpoint saving name

SamitHuang:
910b bugs (on MS master 0919): no more in-graph print

checkpoint saving name: sd-{epoch}.ckpt -> vc_unet-{epoch}.ckpt

gradient accumulation can be set via configs/train_base.py

### DIFF
--- a/examples/videocomposer/configs/train.yaml
+++ b/examples/videocomposer/configs/train.yaml
@@ -16,4 +16,4 @@ epochs: 50
 
 # accelerate or memory reduction
 dataset_sink_mode: True
-use_recompute: True
+use_recompute: False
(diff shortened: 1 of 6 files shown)


examples/videocomposer/configs/train.yaml
```diff
@@ -16,4 +16,4 @@ epochs: 50
 
 # accelerate or memory reduction
 dataset_sink_mode: True
-use_recompute: True
+use_recompute: False
```
wtomin, line 19: Why not using recompute by default?

hadipash (approved):

examples/videocomposer/configs/train.yaml
SamitHuang: It scarifices speed for more memory. If 910B memory is enough, than it's not necessary by default.

PR #155, opened 2023-09-21
tango train

genshimamber:


examples/tango/modules/train/optim.py
hadipash, line 1: We should extract common modules rather than copy-paste them, otherwise it will be a disaster to maintain all these copies. For example, `optim.py` already doesn't match `examples/stable_diffusion_v2/ldm/modules/train/optim.py` even though a week ago it did.

genshimamber: you're absolutely right, although the initial plan was to get the functionalities out first, then we factor these common modules into /mindone (which is planned to happen in Q4, since downstream customers are not aware of bottom level implementations).

hadipash: How about using `from examples.stable_diffusion_v2.ldm.modules.train.optim import build_optimizer` directly to reduce workload during the refactoring stage? Since the details of implementation are of lower priority now.

genshimamber: Thank you for the suggestion. Refactoring usually contains three stages: 1. add common module without changing existing dependencies; 2. change dependencies (in this case, the imports). 3. delete old modules. The 2nd stage is test-heavy, and changing it before stage1 would likely lead to doubled workload.

PR #157, opened 2023-09-22
accelerate T2I-Adapter inference

hadipash:
Moved all operations on tensors (multiplication, concatenation, etc.) under `construct()` to improve inference speed.


SamitHuang:
How is the inference speed improved quantitatively?

hadipash:
> How is the inference speed improved quantitatively?

For batch size of 5: from 220 s/batch down to 22 s/batch (on 910B). I think only partial speed improvement is coming from these changes, the biggest improvement is coming from allowing the network to run in the full graph mode.

SamitHuang (approved):

PR #165, opened 2023-09-29
correct cfg.style_image no attribute error

wtomin:
Since `vc/config/parser.py` has changed in #159 , any argument without a default value will be treated as None, which is not going to be updated in `cfg`. This can cause ` cfg.style_image no attribute error` in `vc/engine/inference_single.py`. Like `--sketch_path` an d `--image_path`, `--style_image` should have a default value.

### DIFF
--- a/examples/videocomposer/vc/config/parser.py
+++ b/examples/videocomposer/vc/config/parser.py
@@ -76,7 +76,7 @@ class Config(object):
         parser.add_argument(
             "--style_image",
             default="",
-            help="Single Sketch Input",
+            help="Style image input",
             type=str,
         )
         parser.add_argument(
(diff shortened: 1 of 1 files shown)


zhtmike (approved):

examples/videocomposer/vc/config/parser.py
```diff
@@ -75,6 +75,7 @@ def _parse_args(self):
         )
         parser.add_argument(
             "--style_image",
+            default="",
```
hadipash, line 78: After setting `style_image` to an empty string, will `read_image_if_provided` fail because it can't open an empty path with `img = Image.open(open(path, mode="rb")).convert("RGB")`?

SamitHuang: should set None as default, so that it can be overwritten by yaml config arg.

SamitHuang (approved):

PR #166, opened 2023-10-03
fix graph inference in fp16 mode

zhtmike:


wtomin (approved):
It does work on 910B ms2.2.0. 
It seems the major difference before and after this commit is :
1. Before this commit: `GroupNorm.to_float(ms.float32)  # changes x.dtype and param.dtype to fp32 before calling construct, will return fp32 out`
2. After this commit:`GroupNorm  # changes x.dtype to fp32 and out.dtype to x.dtype inside construct, will return fp16 if x is fp16`.

I'm wondering whether the reason of the graph inference bug is the overflow of GroupNorm computation. You see in both 1. and 2, groupnorm computation is running under fp32, not fp16. To me, it isn't overflow under fp16 data type that causes this graph inference bug...

SamitHuang (approved):

PR #167, opened 2023-10-04
Change some operators to fp32 calculation implicitly 

zhtmike:
Change some operators to fp32 calculation implicitly instead, for preventing certain bug (e.g, `to_float` in Conv3d and GroupNorm)


SamitHuang:
These changes need to be verified for training in the future. e.g. how is the performance degraded after fp32 conversion,  does it benefit the training outcome?

SamitHuang (approved):

SamitHuang (approved):

PR #175, opened 2023-10-09
videocomposer: change relative import .. in dataset

wtomin:
A minor change to avoid the error of not finding `annotator` when `export MS_COMPILER_CACHE_ENABLE=1`. This bug will be fixed in future mindspore version.

### DIFF
--- a/examples/videocomposer/vc/data/dataset_train.py
+++ b/examples/videocomposer/vc/data/dataset_train.py
@@ -6,11 +6,11 @@ import cv2
 import numpy as np
 import pandas as pd
 from PIL import Image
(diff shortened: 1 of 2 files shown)


HaoyangLee:
Does `export MS_COMPILER_CACHE_ENABLE=1` also work in the inference part?
https://github.com/mindspore-lab/mindone/blob/master/examples/videocomposer/vc/infer_engine/modules/extractor.py#L11
https://github.com/mindspore-lab/mindone/blob/master/examples/videocomposer/deploy/data/datasets.py#L10

zhtmike (approved):

HaoyangLee (approved):

PR #178, opened 2023-10-11
Test videocomposer two steps [FOR DEBUG]

wtomin:
- [x] control randomness of inputs: remove randomresize from misc_transforms; change start_index to deterministic index; use only one video as the training sample.
- [x] control randomness of model: change dropout to zero. VAE encode deterministic
- [x] save ckpt for every step(epoch, since one epoch has one step);
Running command:
```
bash run_train_two_steps.sh
``` 
Change `--ms_mode ` in `run_train_two_steps.sh`, graph mode takes too long. Now I use Pynative mode.

This will save some intermediate `.npy` and `.pkl` during the forward pass in order to align the intermediate results with torch_npu. Checking gradients is not relevant to these files, so you may ignore them.


SamitHuang:
This PR was important for localize the precision issue of VC. But it's not appropriate to merge into the main model code. What about leave it in personal repo if there is no time to wrap into an independent debug script?

wtomin:
Of course. I did not intend to merge this pr into the main branch. It was for other colleagues to take as a reference. I am about to close this PR.

PR #181, opened 2023-10-11
Videocomposer: update CenterCrop for alignment

wtomin:
Change the center crop function in ms transforms to achieve better alignment with torch transforms. This commit is verified to work properly with the training script.

@zhtmike  need to verify and change the ms transforms in infer and deploy folders as well.


SamitHuang (approved):

SamitHuang:
Should also sync to transform in inference phase in `vc/infer_engine/modules/builder.py`

PR #189, opened 2023-10-13
[VC] Skip broken data and select valid one while training, raise error while inferring

HaoyangLee:
- Training: skip broken data and randomly select a valid one (instead of training a dummy video) to achieve better accuracy.
- Inference: raise error instead of inferring a dummy video.


zhtmike (approved):

SamitHuang:
closed due to already being implemented

PR #190, opened 2023-10-13
910B parallel training and finetune ok version: rm assert and enable REF mode in infer

SamitHuang:
Env: 910B + CANN 7.0-T2 + MS2.2-0907 version

cmd: 
   bash run_train_distribute.sh
  
Key setting:
   recompute: ON

Note that: theses change may not be the reason leading to the successful run on 910B. The correct CANN version and MS2.2 can be the real game changer.


zhtmike (approved):

CaitinZhao (approved):

HaoyangLee (approved):

kingcong:
231013单机8卡MindSpore 0907daily + CANN 7.0.T2基线版本代码

PR #205, opened 2023-10-30
Fix optim global step for ms2.2

SamitHuang:
refer to: https://gitee.com/mindspore/mindspore/pulls/57536/files 

Make learning rate update as expected for self-define optimizers.

This bug exists for MS version>=2.2 on both **910A and 910B**


wtomin (approved):

SamitHuang:
close due to adamw_fp32 invoke ms.nn.optim.adamw, which already do global_step+1 after get_lr()

PR #206, opened 2023-10-31
fix Flash Attention

hadipash:
Flash Attention requires the head dimension to be divisible by 16. At the same time, `q` and `k` arrays have `(b * h, n, d)` shape, where the head dimension `d` is at the second axis. This PR fixes the FA requirement check.


hadipash:
This is a limitation of MindSpore's interaface:
https://gitee.com/mindspore/mindspore/blob/b8d67c160fd2e7c7365aa34301d6213e0c06a5a8/mindspore/python/mindspore/nn/layer/flash_attention.py#L221

PR #210, opened 2023-11-02
Support UNCLIP training

zhtmike:


SamitHuang (approved):
LGTM

examples/stable_diffusion_v2/common.py
```diff
@@ -46,8 +46,14 @@ def init_env(
...
     if distributed:
-        init()
         device_id = int(os.getenv("DEVICE_ID"))
+        ms.set_context(
+            mode=mode,
+            device_target="Ascend",
+            device_id=device_id,
+            ascend_config={"precision_mode": "allow_fp32_to_fp16"},  # Only effective on Ascend 901B
+        )
+        init()
```
hadipash, line 56: Why must `set_context` go before `init`?
P.S. Can move `device_id = int(os.getenv("DEVICE_ID"))` and `ms.set_context()` on top to reduce code duplication.

hadipash (approved):

examples/stable_diffusion_v2/common.py
zhtmike: for MS2.1, the default mode change to pynative. So when start the training, the pipeline is "init pynative mode -> Init NCCL in pynative mode -> change to graph mode" will cause error in MS2.1, and it has no problem in MS2.0 since the default mode is graph mode.

wtomin (approved):

PR #216, opened 2023-11-10
add train dreambooth sd 1.5 yaml

Songyuanwei:


examples/stable_diffusion_v2/configs/train_dreambooth_sd_v1.yaml
hadipash, line 1: As this file is identical to `examples/stable_diffusion_v2/configs/v1-train.yaml`, would it be better to specify the only different parameter:
```yaml
model:
  target: ldm.models.diffusion.ddpm.LatentDiffusion
```
And then use `Omega.merge()`? In this case, it would be easy to see how DreabBooth is different from the regular SD.

SamitHuang (approved):

PR #219, opened 2023-11-22
sdv1.5 layernorm_eps changed to 1e-7

wtomin:


SamitHuang:
To avoid making this mistake again, we can change the default eps in LayerNorm to 1e-5?

SamitHuang (approved):
LGTM. Please fix the format for merging

zhtmike (approved):

PR #228, opened 2023-12-01
support IP Adapter (inference only)

zhtmike:
TODO: training pipeline + README


SamitHuang (approved):

examples/stable_diffusion_xl/gm/modules/embedders/modules.py
```diff
@@ -181,7 +181,13 @@ def embedding(self, *tokens, force_zero_embeddings=None):
...
                         crossattn = emb
                     else:
-                        crossattn = ops.concat((crossattn, emb), 2)
+                        if crossattn.shape[1] == emb.shape[1]:
+                            crossattn = ops.concat((crossattn, emb), 2)
+                        else:
+                            # for image/text emb fusion
+                            if emb.shape[0] == 1:
+                                emb = ops.tile(emb, (crossattn.shape[0], 1, 1))
+                            crossattn = ops.concat((crossattn, emb), 1)
```
zhanghuiyao, line 190: Is it certain that image_emb.shape[1] will not be equal to crossattn.shape[1]?

zhanghuiyao (approved):

examples/stable_diffusion_xl/gm/modules/embedders/modules.py
zhtmike: the `if emb.shape[0] == 1:` branch will only be touched for unconditioanl guidance of image embedding, in this case the a zero-value image embedding will be forced replaced with batch size = 1 (since we cannot determine the batch size for some code structure limitation)

PR #239, opened 2023-12-09
[feat] Support PanGu Draw 3.0

townwish4git:
add PanGu Draw 3.0 model below mindone/examples


SamitHuang:
It seems a large amount of code is duplicated with sdxl and hard to see the increased or different parts. It would be better to add illustration for it, e.g. in the PR page

townwish4git:
[https://github.com/mindspore-lab/mindone/compare/master...townwish4git:mindone:master]
This is a version of pangu based on a direct modification of sdxl, all the components needed for the pangu model are added directly to the sdxl code, perhaps this will make the incremental part of pangu clearer

PR #246, opened 2023-12-13
fixed bug on ms2.1

Songyuanwei:
sdxl lora finetune can be run on 910a with ms2.1 or ms 2.2.10


examples/stable_diffusion_xl/gm/models/trainer_factory.py
```diff
@@ -157,10 +157,10 @@ def __init__(
         self.clip_norm = clip_norm
 
         self.enable_accum_grad = False
+        self.accum_step = ms.Parameter(ms.Tensor(0, dtype=ms.int32), name="accum_step")
         if grad_accum_steps > 1:
             self.enable_accum_grad = True
             self.accum_steps = grad_accum_steps
-            self.accum_step = ms.Parameter(ms.Tensor(0, dtype=ms.int32), name="accum_step")
```
zhanghuiyao, line 163: Why move outside？

Songyuanwei: 无标题
A strange error when `self.accum_step=ms.Parameter() ` is inside in ms2.1,  but it's ok in ms2.2. So i move it outside.

wtomin: Referring to `TrainOneStepWrapper` on sdv2, maybe the parameters related gradient accumulation should be set `requires_grad=False`.  I'm not sure whether it is a critical change, but put it here as a recommendation.

wtomin: I have revised this class to solve this error on ms2.1 910A. Maybe you can take it as a [reference](https://github.com/mindspore-lab/mindone/commit/cd760c5112d90d83b499ec0906eb067d138c8baa). The major revision is listed below:
- change the conditions in `init` and `construct` funcs to whether `self.grad_accum_steps > 1`.
- set `self.accum_step.requires_grad` to `False`.
- set `self.accum_step` to zero after a parameter update to avoid the overflow of  `self.accum_step`.

I have another suggestion, that is to disentangle the `LDMWithLoss` module and `TrainOneStepCell` module. Therefore, you can reuse the `TrainOneStepWrapper` in `sdv2`, which has been proved effective. `TrainOneStepWrapper` in sdv2 can replace the `LatentDiffusionWithLossGrad` in sd-xl.

Songyuanwei: thanks， fixed

wtomin (approved):

SamitHuang (approved):

PR #262, opened 2023-12-20
SD-XL: Textual Inversion Train and Inference

wtomin:
The training and inference scripts for textual inversion. Supports:
- [x] training and inference on 910B(ms2.2) and 910A(ms2.1)
- [x] readme contains two example datasets: cat-toy and Chinese-art;

Some critical modifications that may affect other functions:
- `demo/sampling_without_streamlit.py`;
- `gm/models/trainer_factory.py`; (this is to fix a bug related to gradient accumulation)
- `gm/modules/embedders/*`;

<del> BTW, `train_textual_inversion_new.py` and `gm/models/trainer_factory_new.py` are supplementary files. I initially wanted to disentangle the LDMWithLoss Class from the TrainOneStep Class so they could be independently developed. Later I found it was too difficult because there are too many Classes in `gm/models/trainer_factory.py` now.

`train_textual_inversion_new.py` and `gm/models/trainer_factory_new.py` have been deleted because single-graph implementation may introduce more risks of compilation errors.


examples/stable_diffusion_xl/gm/models/trainer_factory_new.py
SamitHuang, line 1: redundant file?

wtomin: Yes and No. 

When I implemented textual inversions, SD-XL did not support training the conditioner. I wrote `train_textual_inversion_new.py` and `gm/models/trainer_factory_new.py` to support training the conditioners, taking `examples/stable_diffusion_v2` as a reference, and wrote the `LatentDiffusionWithLoss` class that can backpropagate gradients through conditioners.

Later I found Huiyao supported training the conditioner of SD-XL in his recent PR. I wrote `train_textual_inversion.py` based on his PR. 

The reason I kept these two redundant files is to encourage the discussion of disentangling `LatentDiffusionWithLoss` from `TrainOneStepCell` in `gm/models/trainer_factory.py`. I think keeping the two parts independent will be easier for future implementation. `gm/models/trainer_factory_new.py` is a reference file and **will be deleted before merging**.

As for the training speed of `train_textual_inversion_new.py` (referred to as `new`) and  `train_textual_inversion.py` (referred to as `old`), on 910A, `old` was about 50ms faster than `new` for each step (the step time was around 4000ms). On 910B, their time difference was negligible. The convergence speeds of the two scripts were similar and the generated images were similar too.

examples/stable_diffusion_xl/gm/util/util.py
```diff
@@ -105,14 +105,23 @@ def auto_mixed_precision(network, amp_level="O0"):
...
-        network = _auto_black_list(
-            network,
-            AMP_BLACK_LIST
-            + [
-                nn.GroupNorm,
-            ],
-            ms.float16,
-        )
+        try:
+            _auto_black_list(
```
hadipash, line 109: I was also working on this feature in parallel with you. This entire function can be simply replaced with this:
```python
    if amp_level != "O3":
        network = amp.auto_mixed_precision(network, amp_level)
    else:
        network = amp.custom_mixed_precision(network, black_list=amp.AMP_BLACK_LIST + [nn.GroupNorm])

    if amp_level in ["O2", "O3"]:   # unwrap the network from the `_OutputTo32` wrapper
        network = network._backbone

    return network
```
And this will work on both MS2.1 and 2.2.

examples/stable_diffusion_xl/gm/models/trainer_factory.py
```diff
@@ -153,31 +153,30 @@ def __init__(
...
             )
+            loss = F.depend(loss, ops.assign_add(self.accum_step, ms.Tensor(1, ms.int32)))
             if self.accum_step % self.accum_steps == 0:
                 if self.clip_grad:
                     grads = ops.clip_by_global_norm(self.accumulated_grads, self.clip_norm)
                     loss = F.depend(loss, self.optimizer(grads))
                 else:
                     loss = F.depend(loss, self.optimizer(self.accumulated_grads))
                 loss = F.depend(loss, self.hyper_map(F.partial(_grad_clear_op), self.accumulated_grads))
+                loss = F.depend(loss, ops.assign(self.accum_step, ms.Tensor(0, ms.int32)))
```
hadipash, line 179: Setting `self.accum_step` to `0` will turn off gradient accumulation due to the `if not self.accum_steps > 1` condition. I think this statement can be omitted.

examples/stable_diffusion_xl/gm/models/trainer_factory.py
```diff
@@ -153,31 +153,30 @@ def __init__(
...
-            self.accum_step = ms.Parameter(ms.Tensor(0, dtype=ms.int32), name="accum_step")
+        self.accum_steps = grad_accum_steps
+        if self.accum_steps > 1:
+            self.accum_step = ms.Parameter(ms.Tensor(0, dtype=ms.int32), name="accum_step", requires_grad=False)
             self.accumulated_grads = optimizer.parameters.clone(prefix="accum_grad", init="zeros")
             self.hyper_map = ops.HyperMap()
 
     def do_optim(self, loss, grads):
-        if not self.enable_accum_grad:
+        if not self.accum_steps > 1:
```
hadipash, line 163:
```suggestion
        if self.accum_steps <= 1:
```

examples/stable_diffusion_xl/gm/modules/embedders/clip/modeling_clip.py
```diff
@@ -334,6 +334,7 @@ class CLIPTextModel(nn.Cell):
     def __init__(self, config_path, weight=None):
         config = self.from_pretrained(config_path)  # CLIPTextConfig
         super().__init__(config)
```
hadipash, line 336: Passing `CLIPTextConfig` to `nn.Cell`?

examples/stable_diffusion_xl/gm/modules/embedders/clip/modeling_clip.py
```diff
@@ -343,6 +344,81 @@ def get_input_embeddings(self) -> nn.Cell:
...
+        Return:
+            `torch.nn.Embedding`: Pointer to the input tokens Embeddings Module of the model.
+        """
+        model_embeds = self._resize_token_embeddings(new_num_tokens)
+        if new_num_tokens is None:
+            return model_embeds
+
+        # Update base model and current model config
+        self.config.vocab_size = model_embeds.embedding_table.shape[0]
+        self.vocab_size = model_embeds.embedding_table.shape[0]
```
hadipash, line 366: Please check here, it looks like `CLIPTextModel` doesn't have `self.config` and `self.vocab_size` members.

hadipash (commented):
Thanks for your effort in disentangling the project's code. Hope we can use it in the future. 🙂

examples/stable_diffusion_xl/gm/util/util.py
hadipash: I think we should request a change from the MS team to add an optional parameter for `fp32` wrapping. Sometimes, we need just `fp16` output.

examples/stable_diffusion_xl/gm/models/trainer_factory.py
wtomin: No worry. `self.accum_steps` is a constant scalar representing the the number of gradient accumulation steps, e.g., 4. `self.accum_step` is a ms.Tensor that records the current accumulated steps.

examples/stable_diffusion_xl/gm/util/util.py
wtomin: I suppose the `amp.AMP_BLACK_LIST + [nn.GroupNorm]` is the black list when amp_level equals to `O2`, not `O3`, right? Besides, I think we often want fp32 outputs to compute the loss value in full precision, because the loss function often contains softmax, log functions that prefer fp32 format.

examples/stable_diffusion_xl/gm/modules/embedders/clip/modeling_clip.py
wtomin: I think it's not necessary. I guess when someone translated the torch codes into mindspore codes, the torch `CLIPTextModel` inherited from a parent class `PreTrainedModel`, like [this](https://github.com/huggingface/transformers/blob/main/src/transformers/models/clip/modeling_clip.py#L751). BTW, in `PreTrainedModel`, `self.config` is [initiated](https://github.com/huggingface/transformers/blob/e268d7e5dcfc3a2f070114795592853ab4408473/src/transformers/modeling_utils.py#L1193C9-L1193C29)

examples/stable_diffusion_xl/gm/modules/embedders/clip/modeling_clip.py
wtomin: `CLIPTextModel` has `self.config` because of line 337 that I edited. 
It does not have `self.vocab_size` until `self.resize_token_embeddings` is called. 
It makes sense to adjust the `vocab_size` after resizing, so I edited here.

hadipash (approved):

examples/stable_diffusion_xl/gm/util/util.py
hadipash: Blacklist is the list of layers that should be executed in full precision when amp level is `O3`.
Whitelist is the list of layers that should be executed in half precision when amp level is `O2`.

On `fp32` loss calculation better to ask @zhanghuiyao. As I understand the loss is part of the network to which AMP is applied:
https://github.com/mindspore-lab/mindone/blob/96ce7cbacc5f8ddcaa5bf293099e5f8732d82916/examples/stable_diffusion_xl/train.py#L156-L166
Therefore, the precision of the output doesn't affect the loss calculation.

hadipash: Yes, misspelled. Should be `if amp_level != "O2"`.

On `fp32` loss calculation better to ask @zhanghuiyao. As I understand the loss is part of the network to which AMP is applied:
https://github.com/mindspore-lab/mindone/blob/96ce7cbacc5f8ddcaa5bf293099e5f8732d82916/examples/stable_diffusion_xl/train.py#L156-L166
Therefore, the precision of the output doesn't affect the loss calculation in this particular case.

SamitHuang (approved):
LGTM

examples/stable_diffusion_xl/gm/models/trainer_factory_new.py
SamitHuang: I see. `trainer_factory_new.py` put vae and clip inside diffusion core graph, similar to our implementation in SD v2. Actually, @Gaohan123 was working on this single-graph attempt for distributing training too. We can leave it as an option.

SamitHuang: A suggestion for naming, `trainer_factory_v2.py` will be better.

examples/stable_diffusion_xl/gm/models/trainer_factory_new.py
```diff
@@ -0,0 +1,77 @@
...
+    def construct(self, x, *tokens):
+        # get latent target
+        x = self.get_first_stage_encoding(x)
+
+        # get condition
+        vector, crossattn, concat = self.conditioner(*tokens)
+        context, y = crossattn, vector
+
+        # get noise and sigma
+        sigmas = self.sigma_sampler(x.shape[0])
```
SamitHuang, line 46: sigma sampling can be placed outside the graph later

hadipash: But sampling performs ops() on tensors. Is it good to place it outside of `construct`?

examples/stable_diffusion_xl/gm/util/util.py
zhanghuiyao: The loss function calculation is included in `TrainOneStepCell`, and the `amp`  setting will take effect on the loss.

examples/stable_diffusion_xl/gm/models/trainer_factory_new.py
```diff
@@ -0,0 +1,77 @@
+# This file only applies to static graph mode
```
zhanghuiyao, line 1: Why add `trainer_factory_new.py`

wtomin: I keep it to encourage the discussion of disentangling the LDMWithLoss and TrainOneStepCell Class so that the development of the two modules will be more independent. Check my previous comment here:
> When I implemented textual inversions, SD-XL did not support training the conditioner. I wrote `train_textual_inversion_new.py` and `gm/models/trainer_factory_new.py` to support training the conditioners, taking `examples/stable_diffusion_v2` as a reference, and wrote the `LatentDiffusionWithLoss` class that can backpropagate gradients through conditioners.
> 
> Later I found Huiyao supported training the conditioner of SD-XL in his recent PR. I wrote `train_textual_inversion.py` based on his PR.
> 
> The reason I kept these two redundant files is to encourage the discussion of disentangling `LatentDiffusionWithLoss` from `TrainOneStepCell` in `gm/models/trainer_factory.py`. I think keeping the two parts independent will be easier for future implementation. `gm/models/trainer_factory_new.py` is a reference file and **will be deleted before merging**.
> 
> As for the training speed of `train_textual_inversion_new.py` (referred to as `new`) and `train_textual_inversion.py` (referred to as `old`), on 910A, `old` was about 50ms faster than `new` for each step (the step time was around 4000ms). On 910B, their time difference was negligible. The convergence speeds of the two scripts were similar and the generated images were similar too.

I totally understand that the disentanglement will be not easy as it affects so many other modules. The code refactoring might be a future work. But I want to discuss the necessity and workload of it with you.

zhanghuiyao (approved):

examples/stable_diffusion_xl/gm/models/trainer_factory_new.py
wtomin: I have deleted the `trainer_factory_new.py` and `train_textual_inversion_new.py` because Huiyao mentioned that the compilation of a single graph may involve more unexpected errors than the current multi-graph implementation. There is no reason to sacrifice usability for implementation convenience, so I deleted the two files.

examples/stable_diffusion_xl/gm/models/trainer_factory_new.py
wtomin: The two files have been deleted.

PR #265, opened 2023-12-22
sd v2: fixed fid bug for scipy version

Songyuanwei:
when scipy vesion is 1.11.*, fid have a error:  `ValueError: Imaginary component`. Because the `scipy.linalg.sqrtm` interface has changed from 1.10 to 1.11,  we change the way to calculate the square root of the matrix.

### DIFF
--- a/examples/stable_diffusion_v2/requirements.txt
+++ b/examples/stable_diffusion_v2/requirements.txt
@@ -9,5 +9,4 @@ imagesize
 toolz
 pillow==9.1.1
 tqdm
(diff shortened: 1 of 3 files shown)


zhtmike (approved):

SamitHuang:
it's better to adapt the code for newer scipy versions. Downgrading scipy may make other packages built on sci-py uncompatible

zhtmike (approved):

SamitHuang (approved):

PR #270, opened 2023-12-29
SDXL: fix distribute vanilla ft bug

zhanghuiyao:


geniuspatrick (approved):

examples/stable_diffusion_xl/demo/sampling.py
```diff
@@ -3,6 +3,9 @@
 import os
 import time
 
+if os.environ.get("MS_PYNATIVE_GE") != "1":
```
SamitHuang, line 6: Should check MS version before setting environment variable in code

examples/stable_diffusion_xl/scripts/run_distribute_vanilla_ft_multi_graph_910b.sh
```diff
@@ -14,6 +14,7 @@ END_DEVICE=$3
 RANK_SIZE=$4
 DATASET_PATH=$5
 
+export HCCL_CONNECT_TIMEOUT=7200
```
SamitHuang, line 17: Be cautious for over-large timeout. Consider setting HCCL_EXEC_TIMEOUT as well to avoid hiding the error info

zhanghuiyao: Running errors will not be hidden due to timeout

examples/stable_diffusion_xl/demo/sampling.py
zhanghuiyao: It is harmless on the current device

SamitHuang (approved):

PR #278, opened 2024-01-08
support IP-Adapter SDv1.5 training & image2image & inpainting inference & SDXL/SD1.5 controlnet inference

zhtmike:
Enhancements:
- Support IP-Adapter + SDv1.5 training pipeline
- Support IP-Adapter + SDv1.5 Image2Image inference pipeline
- Support IP-Adapter + SDv1.5 image in-painting inference pipeline
- Support IP-Adapter + SDv1.5 ControlNet inference pipeline
- Support IP-Adapter + SDXL-base ControlNet inference pipeline
- Print the learning rate and loss scale during the SDv1.5 training process, which are useful in most of the training cases.

Bug Fixes:
- Fix the SDXL-base ControlNet inference pipeline, which originally misses `class_emb`  parameter loading and using, and may downgrade the inference result. 
- Fix the bug that loss scale extracted in SDv1.5 callback is not correctly showed / saved.
- Fix the bug that when using different AMP level for different embedder, there is a concatenation error.


examples/stable_diffusion_v2/ldm/modules/train/trainer.py
```diff
@@ -148,7 +149,13 @@ def construct(self, *inputs):
                 else:
                     # update LR in each gradient step but not optimize net parameter to ensure the LR curve is
                     # consistent
-                    loss = F.depend(loss, self.optimizer.get_lr())  # .get_lr() will make lr step increased by 1
+                    if self.manually_increase_step:
+                        # for MS >= 2.2, `get_lr`` and `global_step += 1` is decoupled
+                        loss = F.depend(
+                            ops.assign_add(self.optimizer.global_step, self.optimizer.global_step_increase_tensor)
```
SamitHuang, line 155: why use `self.optimizer.global_step_increase_tensor` rather than 1 here?

examples/stable_diffusion_v2/ldm/modules/train/callback.py
```diff
@@ -141,7 +135,7 @@ def on_train_step_end(self, run_context):
                     append_dict={
                         "epoch_num": cur_epoch,
                         "cur_step": cur_step,
-                        "loss_scale": loss_scale_manager.get_loss_scale(),
+                        "loss_scale": cb_params.train_network.scale_sense.asnumpy().item(),
```
SamitHuang, line 138: how is the result different using this api?

examples/ip_adapter/train_sd.py
```diff
@@ -0,0 +1,386 @@
+#!/usr/bin/env python
+"""
+Stable diffusion model training/finetuning
```
SamitHuang, line 3: update comment

examples/stable_diffusion_v2/ldm/modules/train/trainer.py
SamitHuang: btw, after increase the global step manually, will the LR and loss scale changes accordingly? Can't see from line152 to 155

examples/ip_adapter/image_to_image_sd.py
```diff
@@ -144,7 +160,12 @@ def main(args):
 
     # set ms context
     device_id = int(os.getenv("DEVICE_ID", 0))
-    ms.set_context(mode=args.ms_mode, device_target="Ascend", device_id=device_id)
+    ms.set_context(
+        mode=args.ms_mode,
+        device_target="Ascend",
+        device_id=device_id,
+        ascend_config=dict(precision_mode="must_keep_origin_dtype"),
```
wtomin, line 167: Why is the precision mode "must_keep_origin_dtype"？ Does the default "force_fp16" cause any artifacts during your experiments?

wtomin (approved):
Nice Job! LGTM.

hadipash (approved):

examples/stable_diffusion_v2/ldm/modules/train/trainer.py
zhtmike: > why use `self.optimizer.global_step_increase_tensor` rather than 1 here?

it follows the internal implementation of Mindspore optimizer, which uses `global_step_increase_tensor` (which is actually just a constant 1), existed in MS2.2

zhtmike: > btw, after increase the global step manually, will the LR and loss scale changes accordingly? Can't see from line152 to 155. If so, we can just delete `get_lr` branch for simplicity?

it is the minimum change for compatibility of MS2.1 and MS2.2. If deleting `get_lr` branch, we need to check the intergrity on MS2.1

examples/stable_diffusion_v2/ldm/modules/train/callback.py
zhtmike: so far `loss_scale_manager.get_loss_scale()` will only return the initial assignment of the loss scale, not return the real value during training process. I think it is a bug.

examples/ip_adapter/image_to_image_sd.py
zhtmike: I have encountered the problem of fp16 overflow on SDXL when using full fp32 mode, which result in color shift in the generated images. For the safetly, I set all training/evaluation scripts in this project to use `"must_keep_origin_dtype" to make sure the network type is what we expected.

examples/ip_adapter/train_sd.py
zhtmike: done

examples/ip_adapter/image_to_image_sd.py
SamitHuang: Does it mean if we don't set `precision_mode`, the generated images will be degraded?

SamitHuang:
The results look great!

examples/ip_adapter/controlnet_sd.py
SamitHuang, line 1: Better align to the naming philosophy as `image_varition_sd.py`, i.e. {task}_{base model}.py. 

Btw, it will much cleaner if we can combine `{task}_sd.py` and `{task}_sdxl` into one script `{task}.py`, and allow users to select the base model in CLI arg. (Since we are not cloning a model, we aim to build super easy-to-use gen. AI kit:)  )

examples/ip_adapter/image_to_image_sd.py
zhtmike: For IPAdpater + SDXL, the result will be degraded without setting the mode.

examples/ip_adapter/tools/merge_ckpts_sdxl.py
```diff
@@ -1,13 +1,90 @@
...
+    content = re.sub(r"(middle_block.[0-9]*).norm1.beta", r"\1.in_layers.0.beta", content)
+    content = re.sub(r"(middle_block.[0-9]*).norm1.gamma", r"\1.in_layers.0.gamma", content)
+    content = re.sub(r"(middle_block.[0-9]*).norm2.beta", r"\1.out_layers.0.beta", content)
+    content = re.sub(r"(middle_block.[0-9]*).norm2.gamma", r"\1.out_layers.0.gamma", content)
+
+    content = replace(content, "conv.bias", "op.bias")
+    content = replace(content, "conv.weight", "op.weight")
+
+    content = replace(content, "add_embedding.linear_1", "label_emb.0.0")
+    content = replace(content, "add_embedding.linear_2", "label_emb.0.2")
```
SamitHuang, line 84: Better to define a name mapping dict and use for-loop to do the replacement.

examples/stable_diffusion_v2/ldm/data/dataset.py
```diff
@@ -253,6 +258,18 @@ def tokenize(self, text):
...
+    def _clip_tokenize(self, texts):
+        batch_encoding = self.tokenizer(
+            texts,
+            truncation=True,
+            max_length=self.tokenizer.context_length,
+            return_length=True,
+            return_overflowing_tokens=False,
+            padding="max_length",
+        )
+        tokens = np.array(batch_encoding["input_ids"], dtype=np.int32)
```
SamitHuang, line 270: I think the dtype should be np.int64?

SamitHuang (approved):

SamitHuang:
Some suggestions:
1. Describe what features are added in the PR descriptions and key TODOs in future. The code changes are of great amount, they definitely deserves bullet marks
2. There are must be intensive hard work behind for making the generated images almost the same as the original ones. And these precious experience are definitely worth sharing and improve other models (such as the loss scale bug, weight decay, sampler alignment). Consider concluding them in the description.

examples/ip_adapter/controlnet_sd.py
zhtmike: `controlnet` is more like a `pipeline` instead, following the naming style in https://github.com/huggingface/diffusers/tree/main/src/diffusers/pipelines/controlnet, which are universal to tasks like depth2image, canny2image, etc, which is a bit special.

I agree that `{task}_sd.py` ant `{task}_sdxl.py` should be put in the same CLI. So far the original CLI in mindone/SDXL is still under development, we can combine them and unify them into the same interface once it is mature enough.

examples/ip_adapter/tools/merge_ckpts_sdxl.py
zhtmike: I agree. It is a temporary converting script for this project only.

examples/stable_diffusion_v2/ldm/data/dataset.py
zhtmike: It is following the implementation in SDXL

zhtmike:
> Some suggestions:
> 
> 1. Describe what features are added in the PR descriptions and key TODOs in future. The code changes are of great amount, they definitely deserves bullet marks
> 2. It must be a hard-work to make the generated images almost the same as the original ones. And these experiences are definitely worth sharing for making other models better (such as the loss scale bug, weight decay, sampler alignment). Consider concluding them in the description.

Done. Added a brief description of what this PR does.

PR #282, opened 2024-01-09
Compel Library: Prompt Re-weighting

wtomin:
Prompt re-weighting is a favorable feature for T2I models. I found [Compel](https://github.com/damian0815/compel) library because of this [article](https://huggingface.co/docs/diffusers/using-diffusers/weighted_prompts) at the diffuser website, and I think it would be great to migrate it from PyTorch to MindSpore. 

The current commits can support prompt re-weighting with a pre-defined syntax for SDv1.5 and SDv2.0(2.1). For examples, please refer to [examples/compel/readme.md](https://github.com/wtomin/mindone/blob/87b6235d70c16b5b230de6d4c606c975a941ea6f/examples/compel/readme.md).

~~However, these commits edit [examples/stable_diffusion_v2/ldm/modules/encoders/modules.py](https://github.com/mindspore-lab/mindone/compare/master...wtomin:mindone:compel#diff-cf1ae9051679bb3dbee6b36cc51395fce8141bbc88e2de748c7fea22102fae66) and [examples/stable_diffusion_v2/ldm/modules/encoders/text_encoder.py](https://github.com/mindspore-lab/mindone/compare/master...wtomin:mindone:compel#diff-1715bc57a374f51a65731d406d5bbe5cedc2ed5db09f52a5b47230fbb522cc15) in order to support `causal_attention_mask` and `attention_mask` simultaneously. I have verified that before and after the commits, the generated images of the SD model are the same using the same random noise. However, these edits may introduce some risks of API mismatch errors in other colleagues' projects, which need to be carefully verified.~~

The current commits keep most changes in `compel` directory, which is less risky.

### DIFF
--- a/examples/compel/text_to_image_compel.py
+++ b/examples/compel/text_to_image_compel.py
@@ -14,8 +14,7 @@ from sdv2.text_encoder import TextEncoder
 
 import mindspore as ms
 
-workspace = os.path.dirname(os.path.abspath(__file__))
-sys.path.append(workspace)
+sys.path.append("../stable_diffusion_v2")
 from ldm.models.clip.simple_tokenizer import get_tokenizer
 from ldm.models.diffusion.ddim import DDIMSampler
 from ldm.models.diffusion.dpm_solver import DPMSolverSampler
@@ -57,16 +56,15 @@ def numpy_to_pil(images):
 
 # To keep Rewrite ldm.modules.encoders.text_encoder from examples/stable_diffusion_v2 untouched
 # write a new text encoder and replace the old one
-
-
 def init_new_text_encoder(config):
     if isinstance(config, str):
         config = OmegaConf.load(config).model
-    params = config.cond_stage_config.params
+    params = config.params.cond_stage_config.params
     dtype = ms.float16 if params.use_fp16 else ms.float32
     context_length = params.context_length
     tokenizer = get_tokenizer(params.tokenizer_name)
     setattr(tokenizer, "context_length", context_length)
+    upcast_attn = params.get("upcast_attn", False)
     transformer = TextEncoder(
         context_length=context_length,
         vocab_size=params.vocab_size,
@@ -77,7 +75,7 @@ def init_new_text_encoder(config):
         epsilon=params.epsilon,
         use_quick_gelu=params.use_quick_gelu,
         dtype=dtype,
-        upcast_attn=params.upcast_attn,
+        upcast_attn=upcast_attn,
     )
     return transformer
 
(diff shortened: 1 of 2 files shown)


examples/stable_diffusion_v2/ldm/modules/encoders/text_encoder.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, d_model, n_head, dtype=ms.float32):
         self.scaling = self.head_dim**-0.5
         self.dtype = dtype
 
-    def construct(self, query, key, value, attn_mask):
+    def construct(self, query, key=None, value=None, causal_attention_mask=None, attention_mask=None):
```
zhtmike, line 40: better to keep the original arguments, inserted new arguments after them, prevent others model unbroken

examples/stable_diffusion_v2/ldm/modules/encoders/text_encoder.py
```diff
@@ -4,16 +4,16 @@
...
-from mindspore import Parameter, Tensor, ops
+from mindspore import Parameter, ops
 from mindspore.common.initializer import TruncatedNormal, initializer
 
 from ._common import LayerNorm, QuickGELU
 
 
 class SequentialCellWithMaskInput(nn.SequentialCell):
-    def construct(self, input_data, mask):
+    def construct(self, input_data, *args, **kwargs):
```
zhtmike, line 14: don't delete `mask` if possible,  prevent others model unbroken

zhtmike:
if possble, use class inheritance and method overloading to handle this text embedding extraction, since this method is not universal, adding more and more arguments will make the algo. hard to read :S

wtomin:
> if possble, use class inheritance to handle this text embedding extraction, since this method is not universal, adding more and more arguments will make the orignal implementation harder to read :S

It seems the most risky changes are within the text_encoder.py. I will see if I can move these changes in to `compel` directory without touching `examples/stable_diffusion_v2/ldm/modules/encoders/text_encoder.py`.

SamitHuang:
Good feature!
Rather than `examples`, it's more reasonable to put it in `mindone/models/compel`

wtomin:
What about moving `compel` to `mindone/tools`? I don't think `compel` is a generative model, like sd or ldm. It's more a like to tool to manipulate the text embedding.

SamitHuang:
> What about moving `compel` to `mindone/tools`? I don't think `compel` is a generative model, like sd or ldm. It's more a like to tool to manipulate the text embedding.

The models in `mindone/models` can be vae, lora, clip, t5, whare are basic models used to build generative models (referring to diffusers). From this view, it's ok to put it there

PR #286, opened 2024-01-10
Add checkpoint deletion feature to SDXL

HaoyangLee:
The sdxl ckpt files are too large (~13GB each) to save in local server especially `args.save_ckpt_interval` is set to be a small value. We can keep the latest `args.max_num_ckpt` ckpt files and delete those old ones. Keep all ckpts when `args.max_num_ckpt` is 0.

### DIFF
--- a/examples/stable_diffusion_xl/gm/helpers.py
+++ b/examples/stable_diffusion_xl/gm/helpers.py
@@ -662,7 +662,7 @@ def delete_checkpoint(ckpt_queue, max_num_ckpt, only_save_lora):
     """
     Only keep the latest `max_num_ckpt` ckpts while training. If max_num_ckpt == 0, keep all ckpts.
     """
-    if max_num_ckpt > 0 and len(ckpt_queue) >= max_num_ckpt:
+    if max_num_ckpt is not None and len(ckpt_queue) >= max_num_ckpt:
         del_ckpt = ckpt_queue.pop(0)
         del_ckpt_lora = _build_lora_ckpt_path(del_ckpt)
         if only_save_lora:
--- a/examples/stable_diffusion_xl/gm/helpers.py
+++ b/examples/stable_diffusion_xl/gm/helpers.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import stat
 from datetime import datetime
 from typing import List, Union
 
@@ -29,8 +28,6 @@ from PIL import Image
 import mindspore as ms
 from mindspore import Tensor, context, nn, ops
 from mindspore.communication.management import get_group_size, get_rank, init
-
-_logger = logging.getLogger(__name__)
 
 
 class BroadCast(nn.Cell):
@@ -676,16 +673,17 @@ def delete_checkpoint(ckpt_queue, max_num_ckpt, only_save_lora):
         for to_del in del_ckpts:
             if os.path.isfile(to_del):
                 try:
-                    os.chmod(to_del, stat.S_IWRITE)
                     os.remove(to_del)
-                    print(
-                        f"INFO: The ckpt file {to_del} is deleted, because the number of ckpt files exceeds the limit {max_num_ckpt}."
+                    logging.debug(
+                        f"The ckpt file {to_del} is deleted, because the number of ckpt files exceeds the limit {max_num_ckpt}."
                     )
                 except OSError as e:
-                    print(e)
-                    print(f"ERROR: Failed to delete the ckpt file {to_del}.")
+                    logging.error(e)
+                    logging.error(f"Failed to delete the ckpt file {to_del}.")
             else:
-                print(f"WARNING: The ckpt file {to_del} to be deleted does not exist.")
+                logging.debug(
+                    f"The ckpt file {to_del} to be deleted doesn't exist. If lora is not used for training, it's normal that lora ckpt doesn't exist."
+                )
 
 
 def get_interactive_image(image) -> Image.Image:
--- a/examples/stable_diffusion_xl/gm/helpers.py
+++ b/examples/stable_diffusion_xl/gm/helpers.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import stat
 from datetime import datetime
 from typing import List, Union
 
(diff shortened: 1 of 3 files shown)


examples/stable_diffusion_xl/gm/helpers.py
```diff
@@ -644,16 +649,45 @@ def save_checkpoint(model, path, only_save_lora=False):
...
+            del_ckpts = [del_ckpt_lora]
+        else:
+            del_ckpts = [del_ckpt, del_ckpt_lora]
+
+        for to_del in del_ckpts:
+            if os.path.isfile(to_del):
+                try:
+                    os.chmod(to_del, stat.S_IWRITE)
+                    os.remove(to_del)
+                    print(
```
hadipash, line 681: Use a logger to print out messages. And this message could be a debug message as it's not important.

examples/stable_diffusion_xl/gm/helpers.py
```diff
@@ -644,16 +649,45 @@ def save_checkpoint(model, path, only_save_lora=False):
...
+        del_ckpt_lora = _build_lora_ckpt_path(del_ckpt)
+        if only_save_lora:
+            del_ckpts = [del_ckpt_lora]
+        else:
+            del_ckpts = [del_ckpt, del_ckpt_lora]
+
+        for to_del in del_ckpts:
+            if os.path.isfile(to_del):
+                try:
+                    os.chmod(to_del, stat.S_IWRITE)
```
hadipash, line 679: Is it necessary to change the mode of the checkpoint file? It was saved by the same script, so it should be writable by default.

hadipash (approved):

examples/stable_diffusion_xl/gm/helpers.py
HaoyangLee: Updated

examples/stable_diffusion_xl/gm/helpers.py
HaoyangLee: Updated

examples/stable_diffusion_xl/train.py
```diff
@@ -52,6 +52,12 @@ def get_parser_train():
     parser.add_argument("--save_path_with_time", type=ast.literal_eval, default=True)
     parser.add_argument("--log_interval", type=int, default=1, help="log interval")
     parser.add_argument("--save_ckpt_interval", type=int, default=1000, help="save ckpt interval")
+    parser.add_argument(
+        "--max_num_ckpt",
```
wtomin, line 56: It makes more sense to set the default value as `None`. If not specified, `None` means to keep all ckpts. If specified, it needs to be a positive integer, like this:
```
if args.max_num_ckpt is not None:
    assert args.max_num_ckpt > 0, "max_num_ckpt must be a positive integer!"
```

examples/stable_diffusion_xl/gm/helpers.py
```diff
@@ -644,16 +646,46 @@ def save_checkpoint(model, path, only_save_lora=False):
...
+        path_lora = _build_lora_ckpt_path(path)
         ms.save_checkpoint(ckpt_lora, path_lora)
         print(f"save lora checkpoint to {path_lora}")
 
 
+def delete_checkpoint(ckpt_queue, max_num_ckpt, only_save_lora):
+    """
+    Only keep the latest `max_num_ckpt` ckpts while training. If max_num_ckpt == 0, keep all ckpts.
+    """
+    if max_num_ckpt > 0 and len(ckpt_queue) >= max_num_ckpt:
```
wtomin, line 665: Then here can be changed to

wtomin (approved):

examples/stable_diffusion_xl/gm/helpers.py
HaoyangLee: Updated

examples/stable_diffusion_xl/train.py
HaoyangLee: Updated

examples/stable_diffusion_xl/gm/helpers.py
```diff
@@ -632,7 +630,11 @@ def perform_save_locally(save_path, samples):
         base_count += 1
 
 
-def save_checkpoint(model, path, only_save_lora=False):
+def _build_lora_ckpt_path(ckpt_path):
+    return ckpt_path[: -len(".ckpt")] + "_lora.ckpt"
```
zhtmike, line 634: hard to read, just to replace by `ckpt_path.replace(".ckpt", "_lora.ckpt"`)

examples/stable_diffusion_xl/gm/helpers.py
```diff
@@ -644,16 +646,46 @@ def save_checkpoint(model, path, only_save_lora=False):
...
+
+        for to_del in del_ckpts:
+            if os.path.isfile(to_del):
+                try:
+                    os.remove(to_del)
+                    logging.debug(
+                        f"The ckpt file {to_del} is deleted, because the number of ckpt files exceeds the limit {max_num_ckpt}."
+                    )
+                except OSError as e:
+                    logging.error(e)
```
zhtmike, line 681: in this case, use `logging.exception` will directly show the tracback

examples/stable_diffusion_xl/train.py
```diff
@@ -194,6 +200,7 @@ def train(args):
         raise ValueError("args.ms_mode value must in [0, 1]")
 
     # 5. Start Training
+    assert args.max_num_ckpt is None or args.max_num_ckpt > 0, "args.max_num_ckpt must be None or a positive integer!"
```
zhtmike, line 203: for best practise, using raise Exception is better than asset

HaoyangLee: Updated

examples/stable_diffusion_xl/gm/helpers.py
HaoyangLee: Updated

examples/stable_diffusion_xl/gm/helpers.py
HaoyangLee: Updated

zhtmike (approved):

PR #291, opened 2024-01-12
FlashAttention: fix compatibility

wtomin:
After the modifications, I have run `text_to_image.py` on both 910A(ms2.1) and 910B(ms2.2) with SDv1.5 and SDv2.0. All the experiments worked without errors, and the generated images are of good quality.

However, the experiments only related to `CrossAttention`. I haven't tested with `CrossFrameAttention`, because I'm not sure when `CrossFrameAttention` is applied.

**PS**: I just found that SDv1.5 does not use flash-attention by default. Therefore the experiments with SDv1.5 were not meaningful.

### DIFF
--- a/examples/stable_diffusion_v2/ldm/modules/attention.py
+++ b/examples/stable_diffusion_v2/ldm/modules/attention.py
@@ -16,16 +16,45 @@ import logging
 
 import numpy as np
 from ldm.util import is_old_ms_version
+from packaging import version
 
 import mindspore as ms
 from mindspore import nn, ops
 from mindspore.common.initializer import initializer
 
-try:
-    from mindspore.nn.layer.flash_attention import FlashAttention
 
-    FLASH_IS_AVAILABLE = True
-    print("flash attention is available.")
+def new_version():
+    return version.parse(ms.__version__) >= version.parse("2.2")
+
+
+try:
+    if new_version():
+        from mindspore.nn.layer.flash_attention import FlashAttention
+
+        FLASH_IS_AVAILABLE = True
+        print("flash attention is available.")
+    else:
+        from mindspore.ops._op_impl._custom_op.flash_attention.flash_attention_impl import get_flash_attention
+
+        # reference to https://arxiv.org/abs/2205.14135
+        class FlashAttention(nn.Cell):
+            def __init__(self):
+                super(FlashAttention, self).__init__()
+                self.flash_attention = get_flash_attention(tiling_stgy_name="sparse")
+
+            def construct(self, q, k, v, attention_mask=None, dropout_mask=None, alibi_mask=None):
+                # ALiBi, reference to https://arxiv.org/abs/2108.12409
+                _, h, Nq, d = q.shape
+                dim_mask = ops.ones((d,), dtype=ms.int8)
+                scale = d**-0.25
+                q = q * scale
+                k = k * scale
+                o, l, m = self.flash_attention(q, k, v, dim_mask, attention_mask, dropout_mask, alibi_mask)
+                return o  # (b, h, n, d)
+
+        FLASH_IS_AVAILABLE = True
+        print("flash attention is available.")
+
 except ImportError:
     FLASH_IS_AVAILABLE = False
     print("flash attention is unavailable.")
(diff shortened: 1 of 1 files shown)


zhtmike (approved):

SamitHuang:
Good.
The generated images by sd1.5+fa on 910b are also ok?

examples/stable_diffusion_v2/ldm/modules/attention.py
```diff
@@ -16,16 +16,45 @@
...
 from mindspore.common.initializer import initializer
 
+
+def new_version():
+    return version.parse(ms.__version__) >= version.parse("2.2")
+
+
 try:
-    from mindspore.nn.layer.flash_attention import FlashAttention
+    if new_version():
```
hadipash, line 31: A new MS version a year from now will be different from the new version today. It is better to hard-code the exact MS version inside the `if` branch as follows: 
```python
from mindone.utils.version_control import MSVersion

if MSVersion >= "2.2.0":
    ...
```

hadipash (approved):

wtomin:
> Good. The generated images by sd1.5+fa on 910b are also ok?

sdv1.5 does not use flashattention by default. Therefore my exps with sdv1.5 were just using vanilla attentions.

examples/stable_diffusion_v2/ldm/modules/attention.py
wtomin: I think inside the `new_version` function that I wrote, the exact version (2.2) has been hard-coded. It is equivalent to your solution, right?

hadipash: Yes, but not exactly. Say, for instance, if a new version 2.3 is released and it has some compatibility issues, then we would need to call the function `new_new_version` or `even_newer_version`. 😁 Hard-coding eliminates this problem and increases readability, as one can directly read the version limitation in the statement.

wtomin: I got your point and it makes sense to me. Updated.

wtomin: After I updated with [3e21d93](https://github.com/mindspore-lab/mindone/pull/291/commits/3e21d93810bd49e95e89419b8b2ae2bb5cc2740a), there is an error for ms2.1 on 910A:
```
RuntimeError: Unsupported func type for Custom operator on CPU, it should be 'hybrid', 'akg', 'pyfunc' or 'aot' or 'julia', but got [tbe] for Custom operator [Custom_flash_attention_b34108071e6ecfbd49a35ff8bb58abbeab1dfa62618a51f566f6c01f2260b47f]
```

I don't know the reason. If I undo [3e21d93](https://github.com/mindspore-lab/mindone/pull/291/commits/3e21d93810bd49e95e89419b8b2ae2bb5cc2740a), only keep [4a69ba3](https://github.com/mindspore-lab/mindone/pull/291/commits/4a69ba36b409ce1bc832b2223b145fd1efd88517), the following command is working fine:
`python text_to_image.py --prompt "a wolf in winter forest" -v "2.0"`

SamitHuang (approved):

examples/stable_diffusion_v2/ldm/modules/attention.py
wtomin: I figured out that the aforementioned error was caused by calling `MSVersion >= "2.2.0"` for multiple times. I will only keep [4a69ba3](https://github.com/mindspore-lab/mindone/pull/291/commits/4a69ba36b409ce1bc832b2223b145fd1efd88517) to avoid this error on 910A.

hadipash: Interestingly, `new_version` is called multiple times as well. I will check and fix `MSVersion` later,  then. Thanks!

wtomin: I found out the reason because if I edited it like this:
```
NEW_VERSION = MSVersion >= "2.2.0"

if NEW_VERSION:
...
```
Then the inference script went well without errors. I found it weird too.

wtomin:
Because I found a better solution to this problem, this PR is closed to avoid confusion.

PR #295, opened 2024-01-16
FlashAttention: Fix Compatibility 

wtomin:
Another solution to fix the compatibility problem of flash-attention layer. It might be a better solution compared with #291  because using `nn.layers.flash_attention` supports head_dim that is not divisible by 16 by padding it to 16*N.

I have tested `text_to_image.py` on:
1. 910A (ms2.1.0)
1.1 sdv2.0 
 :heavy_check_mark: graph mode
 :heavy_check_mark: pynative mode
1.2 sdv1.5
 :heavy_check_mark: graph mode
 :heavy_check_mark: pynative mode
**sdv1.5 succeed because I restrict the `fa_max_head_dim` to 128 instead of 256 on 910A, otherwise there will be OOM error; This value can be passed through `v1-inference.yaml`**.

2 910B (ms 2.2.10.2023.1124)
2.1 sdv1.5
 :heavy_check_mark: graph mode
 :heavy_check_mark: pynative mode
2.2  sdv2.0 
 :heavy_check_mark: graph mode
:x:  pynative mode
```
out = self.flash_attention(
    q.to(ms.float16), k.to(ms.float16), v.to(ms.float16), mask.to(self.fa_mask_dtype)
)
...
Runtime Error: Index 3 is out of bounds 3.
```
I think it is a weird bug, because graph mode has no such an error.

### DIFF
--- a/examples/stable_diffusion_v2/ldm/modules/attention.py
+++ b/examples/stable_diffusion_v2/ldm/modules/attention.py
@@ -16,14 +16,6 @@ import logging
 
 import numpy as np
 from ldm.util import is_old_ms_version
-
-# import os
-# import sys
-# # TODO: remove in future when mindone is ready for install
-# __dir__ = os.path.dirname(os.path.abspath(__file__))
-# mindone_lib_path = os.path.abspath(os.path.join(__dir__, "../../../../"))
-# sys.path.insert(0, mindone_lib_path)
-# from mindone.utils.version_control import is_910a
 from packaging import version
 
 import mindspore as ms
--- a/examples/stable_diffusion_v2/ldm/modules/attention.py
+++ b/examples/stable_diffusion_v2/ldm/modules/attention.py
@@ -156,15 +156,13 @@ class CrossAttention(nn.Cell):
             enable_flash_attention and FLASH_IS_AVAILABLE and (ms.context.get_context("device_target") == "Ascend")
         )
         if self.enable_flash_attention:
+            # TODO: how high_precision affect the training or inference quality
             if version.parse(ms.__version__) < version.parse("2.2"):
-                self.flash_attention = FlashAttention(
-                    head_dim=dim_head, high_precision=True
-                )  # TODO: how high_precision affect the training or inference quality
+                self.flash_attention = FlashAttention(head_dim=dim_head, high_precision=True)
+                self.fa_mask_dtype = ms.float16  # choose_flash_attention_dtype()
             else:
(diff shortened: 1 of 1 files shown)


hadipash:
Close #291 then to avoid merging as it passed reviews?

wtomin:
> Close #291 then to avoid merging as it passed reviews?

Good suggestion!

SamitHuang:
is the generated image on sd1.5 + 910b good?

wtomin:
> is the generated image on sd1.5 + 910b good?

Yes, the quality is good.

hadipash (approved):
Good analysis of FA! But I find it amusing that FA can cause OOM, even though it is supposed to be more memory-efficient compared to vanilla attention.

wtomin:
> Good analysis of FA! But I find it amusing that FA can cause OOM, even though it is supposed to be more memory-efficient compared to vanilla attention.

The OOM caused by FA is not like the OOM caused by a large batch size. As indicated by [flash attention](https://gitee.com/mindspore/mindspore/blob/master/mindspore/python/mindspore/nn/layer/flash_attention.py#L225) in mindspore library, on 910A, it restricts to head dimension to less than 304, otherwise it will cause UB OOM. Although not 100% sure, I guess UB indicates some ultra high-speed NPU memory that is used for exchange data.

examples/stable_diffusion_v2/ldm/modules/attention.py
```diff
@@ -142,15 +144,19 @@ def __init__(
...
         )
         if self.enable_flash_attention:
-            self.flash_attention = FlashAttention(
-                head_dim=dim_head, head_num=heads, high_precision=True
-            )  # TODO: how high_precision affect the training or inference quality
-            self.fa_mask_dtype = ms.uint8  # choose_flash_attention_dtype()
+            # TODO: how high_precision affect the training or inference quality
+            if version.parse(ms.__version__) < version.parse("2.2"):
+                self.flash_attention = FlashAttention(head_dim=dim_head, high_precision=True)
+                self.fa_mask_dtype = ms.float16  # choose_flash_attention_dtype()
```
hadipash, line 156: it [looks like](https://gitee.com/mindspore/mindformers/blob/dev/mindformers/version_control.py#L194) MS2.2.0 requires fp16:
`if version.parse(ms.__version__) <= version.parse("2.2"):`

hadipash (approved):

SamitHuang (approved):

PR #305, opened 2024-01-19
Add optimizer status save

zhanghuiyao:


geniuspatrick (approved):

examples/stable_diffusion_xl/train.py
```diff
@@ -178,6 +184,7 @@ def train(args):
 
             assert args.version == "SDXL-base-1.0", "Only supports sdxl-base."
             assert args.task == "txt2img", "Only supports text2img task."
+            assert args.optimizer_weight is None, "Not supports load optimizer weight."
```
SamitHuang, line 187: why not support loading optimizer state?

zhanghuiyao: save/load optimizer only support with single-graph now

SamitHuang (approved):

PR #310, opened 2024-01-24
Fix sdv1.5 + FlashAttention bug on 910B

wtomin:
As I investigated on 910B, the reason why sdv1.5 + FA generates noisy images is that `ms.nn.layer.flash_attention` (ms version is `2.2.10.20231124`) behaves wrongly when head dimension equals to `160`. Note that the head dimension of sdv1.5 is one of `[40, 80, 160, 320]`.  

There are three versions of solutions:
- [v1](https://github.com/wtomin/mindone/tree/fa_sdv15_v1): we constraint the max head dimension as 128, which only allows head dims `[40, 80]` to use FA, while other head dims use original Attention;
- [v2](https://github.com/wtomin/mindone/tree/fa_sdv15_v2): we pad the head dims from  `[40, 80, 160]` to `[64, 128, 256]`.
- [v3](https://github.com/wtomin/mindone/tree/fa_sdv15_v3): we only pad head dim `160` to `256`.

When running inference with `text_to_image.py` on 910B, I gradually increase the batch size to trigger the OOM error. Here are the results:
| solution | OOM batch size | Not OOM batch size|
| ---| ---| ---|
| v1 | 300| 284|
| v2 | 366 | 360|
| v3 | 384 | 372|

From the perspective of memory efficiency, I suggest using [v3](https://github.com/wtomin/mindone/tree/fa_sdv15_v3).

However, this maybe **a temporary fix**. If mindspore fix this bug in the future, we should remove this fix.

@Songyuanwei , please help to check whether solution [v3](https://github.com/wtomin/mindone/tree/fa_sdv15_v3) has any effect on the sdv1.5 finetuning experiment. Thank you!

### DIFF
--- a/examples/stable_diffusion_v2/ldm/modules/attention.py
+++ b/examples/stable_diffusion_v2/ldm/modules/attention.py
@@ -13,16 +13,16 @@
 # limitations under the License.
 # ============================================================================
 import logging
+import math
 
 import numpy as np
 from ldm.util import is_old_ms_version
 from packaging import version
 
 import mindspore as ms
+import mindspore.numpy as msnp
 from mindspore import nn, ops
 from mindspore.common.initializer import initializer
-import mindspore.numpy as msnp
-import math
 
 try:
     from mindspore.nn.layer.flash_attention import FlashAttention


zhtmike:
I suggest to submit a issue on mindspore gitee, since this bug introduce some heavily impact on using flash attention.

wtomin:
> I suggest to submit a issue on mindspore gitee, since this bug introduce some heavily impact on using flash attention.

I agree to you. I will approach someone in the MindSpore team and report this bug directly, but this temporary fix is still needed as we need to use FA to meet some ddls.

examples/stable_diffusion_v2/ldm/modules/attention.py
```diff
@@ -207,11 +209,19 @@ def construct(self, x, context=None, mask=None):
             v = v.view(v_b, v_n, h, -1).transpose(0, 2, 1, 3)
             if mask is None:
                 mask = ops.zeros((q_b, q_n, q_n), self.fa_mask_dtype)
+            # FIXME: a trick to pad sdv1.5 head dimensions from 160 to 256
+            if head_dim == 160:
+                # pad to 2**n * 64
+                padding_size = 64 * 2 ** math.ceil(math.log(head_dim / 64, 2)) - head_dim
```
hadipash, line 215: I suggest using `ops.ceil` and `ops.log` for better competibility.

hadipash (approved):
LGTM!

zhtmike (approved):

SamitHuang (approved):

PR #311, opened 2024-01-24
add animatediff training pipeline

SamitHuang:
mainly tested on GPU and 910B, training results are as expected for input video data in 256x256, frames=16.

Features added:
- [x] motion module v1&v2 training
- [x] motion lora training
- [x] common utilities for training like lr scheduler are extracted to `mindone` lib
- [x] lora injection are extracted to `mindone` lib

TODOs:
- [ ] Support V3, SDXL
- [ ] Support distributed training in 512x512 in high performance
- [ ] Precision config optimization
- [ ] Improve temporal consistency and spatial fidelity


examples/animatediff/ad/models/diffusion/ddpm.py
```diff
@@ -351,85 +271,144 @@ def decode_first_stage(self, z):
...
-        elif self.parameterization == "velocity":
-            # target = sqrt_alpha_cum * noise - sqrt_one_minus_alpha_prod * x_start
-            target = self.get_velocity(x_start, noise, t)  # TODO: parse train step from randint
         else:
             raise NotImplementedError()
 
-        loss_simple = self.get_loss(model_output, target, mean=False).mean([1, 2, 3])
+        loss = self.mse_mean(target, model_output)
 
+        """
```
wtomin, line 375: `self.compute_loss` and `self.reduce_loss` are commented here and thus never called, but their definitions are not deleted. Is it intended or not?

examples/animatediff/ad/modules/encoders/modules.py
```diff
@@ -47,7 +48,22 @@ def __init__(
...
+            max_length=self.context_length,
+            return_length=True,
+            return_overflowing_tokens=False,
+            padding="max_length",
+        )
+        tokens = ms.Tensor(batch_encoding["input_ids"], ms.int32)
+        return tokens
+
     def tokenize(self, texts):
+        if self.tokenizer_name == "CLIPTokenizer":
```
wtomin, line 64: If `self.tokenize` now returns `np.ndarray` instead of `ms.Tensor`, so does `self._clip_tokenize`.

examples/animatediff/configs/stable_diffusion/v1-inference-mmv1.yaml
```diff
@@ -0,0 +1,89 @@
...
+      params:
+        use_fp16: True
+        tokenizer_name: "BpeTokenizer"
+        context_length: 77
+        vocab_size: 49408
+        output_dim: 768
+        width: 768
+        layers: 12
+        heads: 12
+        epsilon: 1e-7
```
wtomin, line 88: Is this epsilon 1e-5?

examples/animatediff/configs/stable_diffusion/v1-inference-mmv2.yaml
```diff
@@ -0,0 +1,89 @@
...
+      params:
+        use_fp16: True
+        tokenizer_name: "BpeTokenizer"
+        context_length: 77
+        vocab_size: 49408
+        output_dim: 768
+        width: 768
+        layers: 12
+        heads: 12
+        epsilon: 1e-7
```
wtomin, line 88: Is this epsilon 1e-5?

examples/animatediff/ad/models/diffusion/ddpm.py
SamitHuang: not used. leave for method improvement

examples/animatediff/configs/stable_diffusion/v1-inference-mmv1.yaml
SamitHuang: Yes, fixed. thanks for reminding

examples/animatediff/tools/motion_module_convert.py
```diff
@@ -61,12 +64,25 @@ def _param_info_parser(param_info_txt):
...
+        default="v2",
+    )
     parser.add_argument("--target", type=str, default="../models/motion_module")
     args = parser.parse_args()
 
     src_pt_ckpt = args.src
-    torch_names_txt = "./torch_mm_params.txt"
-    ms_names_txt = "./ms_mm_params.txt"
+    print("mm version: ", args.version)
+    if args.version == "v1":
```
HaoyangLee, line 78: Is it correct to load the same params txt whether args.version == v1 or v2 ?

mindone/schedulers/__init__.py
```diff
@@ -0,0 +1 @@
+from .ddim_schedule import DDIMScheduler
```
hadipash, line 1: Looks like `DDIMScheduler` is missing

mindone/models/lora.py
```diff
@@ -0,0 +1,342 @@
...
+from ..utils.config import get_obj_from_str
+from ..utils.version_control import is_old_ms_version
+
+__all__ = [
+    "LoRADenseLayer",
+    "inject_trainable_lora",
+    "make_only_lora_params_trainable",
+    "merge_lora_to_model_weights",
+    "get_lora_params",
+]
```
hadipash, line 20: In this case, I believe it would be better to apply OOP principles to create a simpler and more understandable API. For example:
```python
model = SomeModel(**params)
lora_model = LoRAModel(model, **params)
lora_model.train_lora_only()
lora_model.merge_with_model()
...
```

mindone/models/lora.py
```diff
@@ -0,0 +1,342 @@
...
+
+        # main/orginal linear layer
+        self.linear = nn.Dense(in_features, out_features, has_bias=has_bias).to_float(dtype)
+
+        # side-path/LoRA linear layers, the bias for lora matric should be False
+        self.lora_down = nn.Dense(in_features, rank, has_bias=False).to_float(dtype)
+        self.lora_up = nn.Dense(rank, out_features, has_bias=False).to_float(dtype)
+
+        if is_old_ms_version():
+            self.dropout = nn.Dropout(keep_prob=1 - dropout_p)
```
hadipash, line 57: It's better to use explicit version control (e.g., `MS_VERSION < "2.0"`), because when a newer version of MS is released (e.g. 2.3), it may not be compatible with the current version. To address compatibility issues, we would need to create an additional function like `not_so_old_version()`. To avoid this, explicit versioning is required.

mindone/schedulers/__init__.py
SamitHuang: it's in mindone/schedulers/ddim_schedule.py  Line 68

mindone/utils/amp.py
```diff
@@ -0,0 +1,59 @@
+import mindspore as ms
+from mindspore import nn
+from mindspore.train.amp import AMP_BLACK_LIST, AMP_WHITE_LIST, _auto_black_list, _auto_white_list
+
+
+def auto_mixed_precision(network, amp_level="O0"):
```
hadipash, line 6: It's not good to expose internal functions (those starting with `_`) as they may change in the future without backward compatibility (as we can see here with `_auto_black_list` which requires `try-except` wrapping). A better approach would be:
```python
    if amp_level != "O2":
        network = amp.auto_mixed_precision(network, amp_level)
    else:
        network = amp.custom_mixed_precision(network, black_list=AMP_BLACK_LIST + [nn.GroupNorm])

    return network
```
I would also make it possible to add additional layers to the exception list through the function's parameters.

mindone/models/lora.py
SamitHuang: It can incur parameter name changes if wrap the model in to LoRAModel.

mindone/models/lora.py
```diff
@@ -0,0 +1,342 @@
...
+        activation=None,
+    ):
+        super().__init__()
+
+        if rank > min(in_features, out_features):
+            raise ValueError(f"LoRA rank {rank} must be less or equal than {min(in_features, out_features)}")
+        self.rank = rank
+        self.scale = scale
+        self.dtype = dtype
+
```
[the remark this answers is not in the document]
SamitHuang: There are only dense layers in LoRA. Because we are still tuning the precision configuration for acceptable training results and performance, mannual precision setting for specific modules serves this purpose and may be dropped in future

mindone/models/lora.py
SamitHuang: fixed

mindone/trainers/optim.py
```diff
@@ -0,0 +1,79 @@
...
+    group_params = []
+    if len(decay_params) > 0:
+        group_params.append({"params": decay_params, "weight_decay": weight_decay})  # 1e-6})
+    if len(other_params) > 0:
+        group_params.append({"params": other_params, "weight_decay": 0.0})
+    group_params.append({"order_params": param_optimizer})
+    _logger.info(f"Parameter grouping result: weight decay {len(decay_params)}, no weight decay {len(other_params)}")
+
+    if name.lower() == "adam":
+        OptimCls = Adam
```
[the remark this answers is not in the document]
SamitHuang: fixed

mindone/trainers/recorder.py
```diff
@@ -0,0 +1,59 @@
...
+        measures (Tuple): measurement values corresponding to the metric names
+        """
+        sep = self.sep
+        line = f"{step}{sep}"
+        for i, m in enumerate(measures):
+            if isinstance(m, ms.Tensor):
+                m = m.asnumpy()
+
+            if isinstance(m, float) or isinstance(m, np.float32):
+                line += f"{m:.4f}"
```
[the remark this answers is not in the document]
SamitHuang: fixed

examples/animatediff/ad/modules/encoders/modules.py
SamitHuang: fixed

mindone/models/lora.py
```diff
@@ -0,0 +1,342 @@
...
+            raise TypeError(
+                f"For '{self.cls_name}', the 'activation' must be str or Cell or Primitive, but got "
+                f"{type(activation).__name__}."
+            )
+        self.activation_flag = self.activation is not None
+
+        self._init_weights()
+
+    def _init_weights(self):
+        self.lora_down.weight.set_data(
```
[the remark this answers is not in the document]
SamitHuang: Although it's less concise, the `init_weights` can be kept for external initialization (e.g. for reset and debug)

mindone/utils/amp.py
SamitHuang: Good advance. It takes time to refractor this part. Maybe in next PR

wtomin (approved):

examples/animatediff/ad/data/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+logger = logging.getLogger()
+
+
+def read_gif(gif_path, mode="RGB"):
+    with Image.open(gif_path) as fp:
+        frames = np.array([np.array(frame.convert(mode)) for frame in ImageSequence.Iterator(fp)])
+    return frames
+
+
+def create_video_transforms(h, w, num_frames, interoplation="bicubic", backend="al", use_safer_augment=True):
```
zhtmike, line 24: typo: interoplation

examples/stable_diffusion_v2/ldm/modules/train/callback.py
```diff
@@ -161,12 +185,11 @@ def on_train_step_end(self, run_context):
 
                 self.step_start_time = time.time()
                 _logger.info(
-                    "epoch: %d step: %d, loss is %.3f, average step time (in %d step(s)): %.3f.",
+                    "epoch: %d step: %d, loss is %.6f",
```
zhtmike, line 188: why delete `average step time` ?

mindone/models/lora.py
zhtmike: I think it is fine to wrap the model. if `LoRAModel` is not a cell, and it is just a wrapper, then there is no name changing.

examples/animatediff/ad/data/dataset.py
SamitHuang: fixed

examples/stable_diffusion_v2/ldm/modules/train/callback.py
SamitHuang: It's to make the log cleaner. When dataset sink is ON, it's not used. When OFF, the average step time still flunctuates with default log_interval. So it's better to get the mean value from the recorded result.log

zhtmike (approved):

Gaohan123 (approved):

mindone/models/lora.py
SamitHuang: It is feasible indeed. so the question is whether it is beneficial. In training, we use `inject_trainable_lora`, and in inference, we use `merge_lora_to_model_weights`. They are run in different launches and don't share common attributes. It's a little bit over-encapsulation to put them in class.

examples/stable_diffusion_v2/ldm/modules/train/callback.py
```diff
@@ -156,7 +180,7 @@ def on_train_step_end(self, run_context):
                 self.step_start_time = time.time()
                 if self.record_lr:
                     _logger.info(
-                        "epoch: %d step: %d, lr: %.7f, loss: %.3f, loss scale: %d, average step time (in %d step(s)): %.3f.",
+                        "epoch: %d step: %d, lr: %.7f, loss: %.6f, loss scale: %d, average step time: %.6f.",
```
zhtmike, line 183: since `(in %d step(s))` is deleted, then `self.log_interval` need also be deleted, otherwise it will raise error

PR #26, opened 2023-06-28
Improve LoRA finetune, update trainer & saver, and fix bugs

SamitHuang:
1. Support save/load LoRA param only. 
2. Support LoRA distributed fine-tune. 
3. Add grad clip and ema (trainer & saver updated), needed for good LoRA training results.
4. Impr pokemon results. 
5. Fix bugs in ms1.9

PR #315, opened 2024-01-26
extract environment initialization

hadipash:
This PR introduces the extraction of the training environment initialization, as it is utilized across the project. I have also added a [graph caching feature](https://www.mindspore.cn/tutorials/en/r2.2/advanced/static_graph_expert_programming.html#using-the-compilation-cache), which is exceptionally helpful during the hyperparameters tuning stage. Here are the numbers (tested on SDv2.1 and pokemon-blip dataset):
| Device | Environment | First epoch time without caching | With caching |
|--------|-------------|----------------------------------|--------------|
| 910    | MS2.1       | 15+ minutes                      | 4~ minutes   |
| 910*   | MS2.2.1     | 17+ minutes                      | 5+ minutes   |


zhtmike:
@HaoyangLee the graph cache is fixed now?

HaoyangLee:
> @HaoyangLee the graph cache is fixed now?

Haven't use that feature for a while. Not sure about it in the latest MS version. @hadipash 's testing result looks nice.

zhtmike (approved):

mindone/env/init.py
```diff
@@ -0,0 +1,90 @@
...
+    enable_modelarts: bool = False,
+    num_workers: int = 1,
+    json_data_path: Optional[str] = None,
+) -> Tuple[int, int, int]:
+    """
+    Initialize MindSpore training environment.
+
+    Args:
+        mode: MindSpore execution mode. Default is 0 (ms.GRAPH_MODE).
+        debug: Whether to enable debug mode (forces PyNative mode). Default is False.
```
zhtmike, line 29: better name as debug -> is_debug_mode

mindone/env/init.py
```diff
@@ -0,0 +1,90 @@
...
+from mindspore.communication import get_group_size, get_rank, init
+
+from ..utils.seed import set_random_seed
+from .parallel_config import ParallelConfig
+
+_logger = logging.getLogger(__name__)
+
+
+def init_train_env(
+    mode: int = ms.GRAPH_MODE,
```
wtomin, line 15: Can you also support setting different device_target? such as GPU

mindone/env/init.py
```diff
@@ -0,0 +1,90 @@
...
+        device_id = int(os.getenv("DEVICE_ID"))
+        ms.set_context(
+            mode=mode,
+            device_target="Ascend",
+            device_id=device_id,
+            ascend_config={"precision_mode": "allow_fp32_to_fp16"},  # Only effective on Ascend 910*
+        )
+        init()
+        device_num = get_group_size()
+        ParallelConfig.dp = device_num
```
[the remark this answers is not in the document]
hadipash: It is used in `ParallelTrainOneStepWithLossScaleCell` but PR #26 removed all its usages. I think I can safely remove `ParallelConfig` from the environment initialization.

mindone/env/init.py
hadipash: Done. CPU support is unnecessary, I suppose?

mindone/env/init.py
hadipash: This sounds like a function that should return an output, `is_debug_mode() -> True / False`, rather than a flag.

mindone/env/init.py
wtomin: Yes, CPU does not need to be supported.

wtomin (approved):

PR #321, opened 2024-01-29
SD V1.5: disable Flash-attention by default

wtomin:
Because of the risk of oom on 910A of sdv1.5+ FA, we should change the default setting and leave the option to the user.


Songyuanwei (approved):

hadipash (approved):

SamitHuang (approved):

SamitHuang:
Please resolve the conflict

zhtmike (approved):

PR #326, opened 2024-01-30
Update ControlNet SDXL inference pipeline and document

HaoyangLee:


zhtmike:
Btw, I suggest to move the controlnet condition as one part of the GeneralConditioner in SDXL, as it allows easier integration with controlnet condition estimator, like MiDas. Currently implementation seems not feasible with the neural network estimators.

HaoyangLee:
> Btw, I suggest to move the controlnet condition as one part of the GeneralConditioner in SDXL, as it allows easier integration with controlnet condition estimator, like MiDas. Current implementation seems not feasible with the neural network estimators.

Good suggestion, can make a separate PR later.

SamitHuang (approved):

examples/stable_diffusion_xl/demo/sampling_without_streamlit.py
zhtmike, line 1: move line 189  `control, H, W = get_control(args, num_samples, min(H, W))` before the definition of value_dict (line 162), since the H, W is updated but not updated in `value_dict`.

zhtmike (approved):

examples/stable_diffusion_xl/demo/sampling_without_streamlit.py
HaoyangLee: fixed

PR #363, opened 2024-02-28
Dit model definition

wtomin:
The model definition file’ `mindone/models/dit.py`, with flash-attention support. When enabling FA, the inference results on 910B with ms 2.2.10 were ok.

Usage:
```python
from mindone.models.dit import DiT_models
model = DiT_models["DiT_XL_2"](input_size=32, block_kwargs={"enable_flash_attention": True})
```

TODO:
- [ ] Automatic downloading of pretrained checkpoints

### DIFF
--- a/mindone/models/dit.py
+++ b/mindone/models/dit.py
@@ -4,11 +4,12 @@ import numbers
 from typing import Optional, Tuple
 
 import numpy as np
-from utils.version_control import check_valid_flash_attention, choose_flash_attention_dtype
 
 import mindspore as ms
 from mindspore import Parameter, Tensor, nn, ops
 from mindspore.common.initializer import Constant, Normal, One, XavierNormal, XavierUniform, Zero, initializer
+
+from ..utils.version_control import check_valid_flash_attention, choose_flash_attention_dtype
 
 logger = logging.getLogger(__name__)
 FLASH_IS_AVAILABLE = check_valid_flash_attention()
@@ -16,6 +17,23 @@ if FLASH_IS_AVAILABLE:
     from mindspore.nn.layer.flash_attention import FlashAttention
 
     logger.info("Flash attention is available.")
+
+__all__ = [
+    "DiT",
+    "DiT_models",
+    "DiT_XL_2",
+    "DiT_XL_4",
+    "DiT_XL_8",
+    "DiT_L_2",
+    "DiT_L_4",
+    "DiT_L_8",
+    "DiT_B_2",
+    "DiT_B_4",
+    "DiT_B_8",
+    "DiT_S_2",
+    "DiT_S_4",
+    "DiT_S_8",
+]
 
 
 def normal_(tensor: Tensor, mean: float = 0.0, std: float = 1.0):
--- a/mindone/models/dit.py
+++ b/mindone/models/dit.py
@@ -1,0 +1,691 @@
+import logging
+import math
+import numbers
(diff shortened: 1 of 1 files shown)


geniuspatrick (approved):

mindone/models/dit.py
```diff
@@ -0,0 +1,709 @@
...
+                finfo_type = np.float16
+            else:
+                finfo_type = np.float32
+            max_neg_value = -np.finfo(finfo_type).max
+            mask = mask.repeat(self.heads, axis=0)
+            mask = ops.expand_dims(mask, axis=1)
+            sim.masked_fill(mask, max_neg_value)
+
+        # use fp32 for exponential inside
+        attn = self.softmax(sim.astype(ms.float32)).astype(v.dtype)
```
SamitHuang, line 195: is softmax also computed in fp32 in original impl?

mindone/models/dit.py
```diff
@@ -703,6 +710,263 @@ def DiT_S_8(**kwargs):
...
+        """
+        # https://github.com/openai/glide-text2im/blob/main/notebooks/text2im.ipynb
+        half = x[: len(x) // 2]
+        combined = ops.cat([half, half], axis=0)
+        model_out = self.construct(combined, t, y=y, text_embed=text_embed)
+        # For exact reproducibility reasons, we apply classifier-free guidance on only
+        # three channels by default. The standard approach to cfg applies it to all channels.
+        # This can be done by uncommenting the following line and commenting-out the line following that.
+        # eps, rest = model_out[:, :self.in_channels], model_out[:, self.in_channels:]
+        eps, rest = model_out[:, :, :3], model_out[:, :, 3:]
```
SamitHuang, line 939: why only keeps 3 channels in the original impl?

mindone/models/dit.py
wtomin: In [DiT torch implementation](https://github.com/facebookresearch/DiT/blob/main/models.py), they used `timm.vision_transformer.Attention`. The code implies they are not forcing fp32 for softmax computation: https://github.com/huggingface/pytorch-image-models/blob/main/timm/models/vision_transformer.py#L57

```python
    def forward(self, x: torch.Tensor) -> torch.Tensor:
        B, N, C = x.shape
        qkv = self.qkv(x).reshape(B, N, 3, self.num_heads, self.head_dim).permute(2, 0, 3, 1, 4)
        q, k, v = qkv.unbind(0)
        q, k = self.q_norm(q), self.k_norm(k)

        if self.fused_attn:
            x = F.scaled_dot_product_attention(
                q, k, v,
                dropout_p=self.attn_drop.p if self.training else 0.,
            )
        else:
            q = q * self.scale
            attn = q @ k.transpose(-2, -1)
            attn = attn.softmax(dim=-1)
            attn = self.attn_drop(attn)
            x = attn @ v

        x = x.transpose(1, 2).reshape(B, N, C)
        x = self.proj(x)
        x = self.proj_drop(x)
        return x

```

mindone/models/dit.py
wtomin: There is a PR related to it. [the latent channels should be 4 instead of 3](https://github.com/facebookresearch/DiT/pull/12#top). The DiT author explains that cfg on 4 channels is correct, but for reproducibility reasons, they leaves the wrong code and suggest users to revise it manually. I will correct it soon.

SamitHuang (approved):

PR #364, opened 2024-02-28
SDXL: resume from specified training step

townwish4git:
# What does this PR do?

Adds # (feature)
resume training from a specified training step

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you build and run the code without any errors?

## Who can review?

@xxx

### DIFF
--- a/examples/stable_diffusion_xl/train.py
+++ b/examples/stable_diffusion_xl/train.py
@@ -228,7 +228,7 @@ def train(args):
     scaler = args.rank_size * dataloader.get_batch_size() * args.gradient_accumulation_steps if args.scale_lr else 1.0
     lr = get_learning_rate(config.optim, config.data.total_step, scaler)
     if "scheduler_config" in config.optim and args.resume_step:
-        lr = lr[args.resume_step:]
+        lr = lr[args.resume_step :]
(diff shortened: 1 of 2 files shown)


examples/stable_diffusion_xl/train.py
```diff
@@ -125,6 +125,7 @@ def get_parser_train():
         default=None,
         help="Max number of ckpts saved. If exceeds, delete the oldest one. Set None: keep all ckpts.",
     )
+    parser.add_argument("--resume_step", type=int, default=0, help="resume from step_n")
```
SamitHuang, line 128: I think we should record the step number in the saving checkpoint so that it can be loaded when resume training  instead of manual setting.

zhanghuiyao (approved):

examples/stable_diffusion_xl/train.py
zhanghuiyao: > I think we should record the step number in the saving checkpoint so that it can be loaded when resume training instead of manual setting.

It makes sense that we can continue to optimize in the next PR @townwish4git

geniuspatrick (approved):

PR #395, opened 2024-03-13
Latte: inference scripts and training on Sky TimeLapse

wtomin:
# What does this PR do?

1. The ckpt conversion scripts for vae and latte checkpoint;
2. The inference scripts for latte using sky-timelapse pretrained ckpt and ucf-101 pretrained ckpt;
3. The training dataset for sky timelapse dataset;
4. The training script of latte on the sky timelapse: the loss on Ascend is decreasing. The ckpt at epoch=150 generates reasonable videos.

TODOs:
1. To support more datasets for training;
2. Speed acceleration (Flash-Attention is not working correctly! Needs to be fixed).


examples/latte/README.md
SamitHuang, line 1: can we append some training results as a reference?

examples/dit/diffusion/gaussian_diffusion.py
```diff
@@ -149,7 +149,12 @@ def p_mean_variance(self, model, x, t, clip_denoised=True, denoised_fn=None, mod
         if model_kwargs is None:
             model_kwargs = {}
 
-        B, C = x.shape[:2]
+        if x.dim() == 4:
```
zhtmike, line 152: better to decouple from dit module

wtomin: Fixed.

examples/latte/README.md
wtomin: Training results have not been ready yet.

zhtmike (approved):

hadipash (approved):

mindone/models/latte.py
```diff
@@ -0,0 +1,312 @@
...
+    xavier_uniform_,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class Latte(nn.Cell):
+    def __init__(
+        self,
+        input_size=32,
```
hadipash, line 26: Since this part is being integrated into the library, it would be nice for the code to be well typed and documented.

wtomin: Thanks for your advice. It is improved.

PR #409, opened 2024-03-25
Latte training bug fix: temp_embed initialization

wtomin:
Fix the bug of Latte model initialization: failed to initialize temporal position embedding.


mindone/models/modules/pos_embed.py
```diff
@@ -25,6 +25,14 @@ def get_2d_sincos_pos_embed(embed_dim: int, nh: int, nw: Optional[int] = None) -
     return pos_embed
 
 
+def get_1d_sincos_temp_embed(
```
zhtmike, line 28: add docstring, and type hint of return type

wtomin: Fixed~

zhtmike (approved):

SamitHuang (approved):

CaitinZhao (approved):

PR #413, opened 2024-03-27
[AD] Improve train_resume.ckpt saving to adapt for resuming training automatically on modelarts

HaoyangLee:
# What does this PR do?

Adds # (feature)
Modelarts supports automatic training resume. But the resume ckpt saving method should be improved to adapt this.

Before: `train_resume.ckpt` will save in a different directory formed by a different timestamp after each launch of resuming training. If the training will be resumed multiple times, each time we need to manually modify the `train_resume.ckpt` path (i.e. `args.resume`) to ensure that the LATEST train_resume.ckpt instead of the older one is loaded.

After: `train_resume.ckpt` is always saved in `args.output_path`. Set `args.resume` to be `True` at the first launch of training, then the latest `train_resume.ckpt` can be loaded correctly even if the training will be resumed multiple times, without setting `args.resume` again manually.

## Before submitting
- [x] Did you build and run the code without any errors?
- [x] Did you report the running environment (NPU type/MS version) and performance in the doc? (better record it for data loading, model inference, or training tasks)
- [x] Did you write any new necessary tests?

### DIFF
--- a/examples/animatediff/args_train.py
+++ b/examples/animatediff/args_train.py
@@ -17,6 +17,15 @@ def _check_cfgs_in_parser(cfgs: dict, parser: argparse.ArgumentParser):
             raise KeyError(f"{k} does not exist in ArgumentParser!")
 
 
+def _parse_bool_str(b: str):
+    """Allow input args to be either str2bool or str (e.g. a filepath)."""
+    if b.lower() not in ["false", "true"]:
+        return b
+    if b.lower() in ["false"]:
+        return False
+    return True
+
+
 def parse_args():
     parser = argparse.ArgumentParser()
     parser.add_argument(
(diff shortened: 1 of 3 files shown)


hadipash (approved):

examples/animatediff/args_train.py
```diff
@@ -17,6 +17,15 @@ def _check_cfgs_in_parser(cfgs: dict, parser: argparse.ArgumentParser):
             raise KeyError(f"{k} does not exist in ArgumentParser!")
 
 
+def _parse_bool_str(b: str):
+    """Allow input args to be either str2bool or str (e.g. a filepath)."""
+    if b.lower() not in ["false", "true"]:
+        return b
+    if b.lower() in ["false"]:
+        return False
+    return True
```
hadipash, line 26: How about modifying `mindone.utils.config.str2bool` by adding to it a `return_str` parameter? Something like:
```python
def str2bool(b: str, return_str: bool = False):
    if b.lower() not in ["false", "true"]:
        if return_str:
            return b
        else:
            raise Exception("Invalid Bool Value")
...
```

PR #433, opened 2024-04-09
T5 text encoder inference: graph and pynative mode

wtomin:
# What does this PR do?

Adds # T5 text encoder inference, supporting both pynative and graph mode.
Readme includes:
- [x] how to convert the checkpoint of `t5-v1_1-xxl`;
- [x] how to use the text encoder to extract text embedding and mask;
- [ ] how to use text decoder (todo).


mindone/models/t5/modules/pretrained_config.py
```diff
@@ -0,0 +1,234 @@
...
+import os
+from typing import Dict, Optional, Tuple
+
+from mindspore import log as logger
+
+from .configs import HF_CONFIG_URL_BASE
+from .download import cached_path
+
+
+class PreTrainedConfig:
```
hadipash, line 31: `PreTrainedConfig` here and in `mindone/models/t5/modules/configs.py` are identical. Why is there a need for repetition?

mindone/models/t5/t5.py
```diff
@@ -0,0 +1,219 @@
...
+        use_text_preprocessing=True,
+        model_max_length=120,
+        pretrained_ckpt=None,
+    ):
+        super().__init__()
+        self.use_text_preprocessing = use_text_preprocessing
+        self.cache_dir = cache_dir
+        self.pretrained_ckpt = pretrained_ckpt
+
+        self.tokenizer = get_t5_tokenizer(cache_dir)
```
hadipash, line 48: I think tokenizers and embedders should be separated into different modules since they represent different steps in the pipeline. Furthermore, tokenizers operate on strings rather than tensors, so it's preferable to keep them outside of the `nn.Cell`.

mindone/models/t5/t5.py
```diff
@@ -0,0 +1,219 @@
...
+        + r"\*"
+        + r"]{1,}"
+    )  # noqa
+
+    def __init__(
+        self,
+        cache_dir,
+        use_text_preprocessing=True,
+        model_max_length=120,
+        pretrained_ckpt=None,
```
hadipash, line 41: Since this code will be added to the library, please include type hints.

zhtmike: And docstrings for APIs, thanks

hqkate:
After the discussion with Yufei, the T5 encoder will reuse the modules in mindone.transformers, please refer to  #440

PR #439, opened 2024-04-12
Revise Train Scripts [Latte and DiT]: avoid setting positional embedding to trainable parameters

wtomin:
For DiT and Latte models, the positional embedding tensor is initialized as a Parameter that does not require gradient updates.

However, in the training script, I accidently set all parameters to trainable. It is not desirable for positional embeddings, because they are initialized and then fixed during training.

### DIFF
--- a/mindone/models/latte.py
+++ b/mindone/models/latte.py
@@ -81,8 +81,8 @@ class Latte(nn.Cell):
             )
         num_patches = self.x_embedder.num_patches
         # Will use fixed sin-cos embedding:
-        self.pos_embed = ms.Parameter(ops.zeros((1, num_patches, hidden_size), dtype=ms.float32), requires_grad=False)
-        self.temp_embed = ms.Parameter(ops.zeros((1, num_frames, hidden_size), dtype=ms.float32), requires_grad=False)
+        self.pos_embed = ops.zeros((1, num_patches, hidden_size), dtype=ms.float32)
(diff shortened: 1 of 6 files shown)


zhtmike:
maybe just change to tensor and skip load the pos embed parameter, prevent accident change all parameter to be trainable

wtomin:
> maybe just change to tensor and skip load the pos embed parameter, prevent accident change all parameter to be trainable

Now the pos_embed and temp_embed (in latte only) are Tensors instead of Parameters.

mindone/models/latte.py
```diff
@@ -81,8 +84,8 @@ def __init__(
             )
         num_patches = self.x_embedder.num_patches
         # Will use fixed sin-cos embedding:
-        self.pos_embed = ms.Parameter(ops.zeros((1, num_patches, hidden_size), dtype=ms.float32), requires_grad=False)
-        self.temp_embed = ms.Parameter(ops.zeros((1, num_frames, hidden_size), dtype=ms.float32), requires_grad=False)
+        self.pos_embed = ops.zeros((1, num_patches, hidden_size), dtype=ms.float32)
```
zhtmike, line 87: just change to `self.pos_embed = Tensor(pos_embed).float().unsqueeze(0)` here for easier reading

zhtmike (approved):

hadipash (approved):

SamitHuang (approved):

SamitHuang (approved):

PR #494, opened 2024-05-09
OpenSora v1.1 training script

hadipash:


examples/opensora_hpcai/opensora/models/layers/blocks.py
```diff
@@ -184,10 +185,7 @@ def construct(self, x, cond, mask=None):
         x = ops.reshape(x, (B, N, -1))
 
         # 4. output projection
```
zhtmike, line 187: why change this?

zhtmike (approved):

examples/opensora_hpcai/opensora/models/layers/blocks.py
hadipash: When using Flash Attention, the attention `dtype` changes to bf16, including `self.proj` and `self.proj_drop`. These should be cast back if the network uses a different data type. Otherwise training crash due to different data types.

zhtmike: Then the type conversion should be put after FA instead of the block output?

hadipash: `self.proj` and `self.proj_drop` are also set to `attn_dtype`, so need to cast at the end of the `SelfAttention` and `MultiHeadCrossAttention` blocks.

examples/opensora_hpcai/README.md
```diff
@@ -39,6 +39,10 @@ Videos are downsampled to `.gif` for display. Click for original videos. Prompts
 
 ## 🔆 Features
 
+- 📍 **Open-Sora 1.0** with the following features
```
SamitHuang, line 42: 1.0 -> 1.1

examples/opensora_hpcai/README.md
SamitHuang, line 1: pls add the inference results for v1.1 with the same prompts used in torch.

examples/opensora_hpcai/README.md
SamitHuang: pls update the new table as well

examples/opensora_hpcai/README.md
```diff
@@ -249,6 +254,10 @@ video_folder/part01/vid001.mp4,a cartoon character is walking through
 video_folder/part01/vid002.mp4,a red and white ball with an angry look on its face
 ```
 
+> [!IMPORTANT]
+> OpenSora v1.1 also requires the `length` field, which represents the number of frames,
```
SamitHuang, line 258: can you add instructions on how to get the `length` for the original text-video data csv?

CaitinZhao (approved):

SamitHuang (approved):

PR #500, opened 2024-05-15
hpcai opensora v1.0 speedup and bug fix

SamitHuang:
# What does this PR do?

Fixes # (issue)
- dvm precision issue 
- inference OOM issue in graph mode

Adds # (feature)
- cast and recompute optimization
- bf16 training

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/mindspore-lab/mindone/blob/master/CONTRIBUTING.md)?
- [ ] Did you make sure to update the documentation with your changes? E.g. record bug fixes or new features in `What's New`. Here are the
- [ ] Did you build and run the code without any errors?

## Who can review?

@xxx


examples/opensora_hpcai/opensora/datasets/video_dataset.py
```diff
@@ -0,0 +1,212 @@
...
+        [
+            SmallestMaxSize(max_size=h, interpolation=mapping[interpolation]),
+            CenterCrop(h, w),
+        ],
+    )
+
+    return pixel_transforms
+
+
+class VideoDataset:
```
hadipash, line 34: For efficient data sampling, data fetching and preprocessing should be separated into different pipelines, as one is IO intensive, and the other is CPU intensive. I integrated such pipelines in [OpenSora v1.1 training](https://github.com/mindspore-lab/mindone/pull/494/files#diff-4af26d61026bd16cbca7135edf4e1fa4d75db7d4a8db947aed46d2c27c653710). Maybe we follow this deisgn?

examples/opensora_hpcai/opensora/models/layers/blocks.py
```diff
@@ -287,7 +286,7 @@ def construct(self, x, mask=None):
         # reshape FA output to original attn input format (b n h*d)
         out = out.view(B, N, -1)
 
-        return self.proj_drop(self.proj(out)).to(x_dtype)
+        return self.proj_drop(self.proj(out))
```
hadipash, line 289: This will cause a problem in FP32 training as FA converts attention's type to BF16.

examples/opensora_hpcai/opensora/utils/amp.py
```diff
@@ -0,0 +1,68 @@
...
+    if amp_level == "O0":
+        pass
+    elif amp_level == "O1":
+        if NEW_AUTO_WHITE:
+            return _auto_mixed_precision_rewrite(network, dtype, white_list=AMP_WHITE_LIST)
+        else:
+            return _auto_white_list(network, AMP_WHITE_LIST, dtype=dtype)
+    elif amp_level == "O2":
+        try:
+            _auto_black_list(
```
hadipash, line 53: There's no reason to expose these internal APIs. The whole function can be rewritten as:
```python
if amp_level != "O2":
    return amp.auto_mixed_precision(network, amp_level, dtype=dtype)
else:
    return amp.custom_mixed_precision(network, black_list=amp.AMP_BLACK_LIST + custom_fp32_cells, dtype=dtype)
```

examples/opensora_hpcai/scripts/train.py
```diff
@@ -368,7 +389,11 @@ def main(args):
         ema=ema,
     )
 
-    model = Model(net_with_grads)
+    if args.global_bf16:
+        model = Model(net_with_grads, amp_level="O0")
```
hadipash, line 393: Why is `amp_level` set to `O0` when the training precision is BF16?

SamitHuang: because only layers in fp32 will be converted bf16 when using `allow_mixed_precision_bf16` to enable operations compute in bf16

examples/opensora_hpcai/opensora/datasets/video_dataset.py
SamitHuang: got it. It should be more efficient.  How is performance gain when using `dataloader.map` to do transformation instead? Is frame consistency preserved for random augmentation ?

examples/opensora_hpcai/opensora/utils/amp.py
SamitHuang: I am not sure whether custom_mixed_precision is well verified. error comes out in initial attempt.

examples/opensora_hpcai/opensora/models/layers/blocks.py
SamitHuang: fixed

examples/opensora_hpcai/opensora/datasets/video_dataset.py
hadipash: Unfortunately, frame consistency for random augmentations is not available in MindData yet. But it can be easily fixed by introducing custom transforms, i.e.:
```python
class HorizontalFlip:
    def __init__(self, p: float = 0.5):
        self._p = p

    def __call__(self, frames: np.ndarray) -> np.ndarray:
        if random.random() < self._p:
            return np.flip(frames, axis=-1)     # assuming HWC order
```

On the performance gain: didn't have time to measure, but theoretically, there should be a noticeable speed boost as videos can be read and processed in parallel rather than in sequence as it is done now (each worker must load a video first).

CaitinZhao (approved):

HaoyangLee (approved):

geniuspatrick (approved):

PR #605, opened 2024-07-23
feat!: Dynamic shape training in graph mode for hpcai-OpenSora v1.1 with more fixes

SamitHuang:
# What does this PR do?

Fixes and updates:
1. Save ema and non-ema checkpoints independently
2. Fix T5 bf16 mixed precision inference in pynative mode, and allow setting T5 dtype
3. Merge dynamic shape adaption for graph mode from @yuedongli1  with some fixes including:
     -  fix "valid" conv for patchify with manual padding
     - rm vae micro batch size
     - set graph fusion flags in init_env
4. Make it more compatible with previous static shape training by
     - only initialize dynamic shape tensors (`set_inputs`) when bucket configs are provided
     - allow setting `manual_pad` for dynamic shape training
     - remove `if True:`, recover `frames_mask` branch after testing to confirm it will NOT increase compilation time, which is **~16 min for stage 2 training**.
     -recover assertions in Cell init for usage reliability

After these fixes and updates, now it can support both i) dynamic shape training with bucket configs from Stage 1 to Stage 3 for jit_level O0, and ii) static shape training in for jit_level O0/O1. Results as follows:
   Env: CANN-C18 (0705)  + MS2.3 official
    **Stage 2 dynamic shape: jit_level=O0, compilation time: ~16 min.   step time:  13s/step to 15 s/step**
    **Static  shape 576x1024x24: jit_level=O1, compilation time: ~11 min.   step time:  7.84 s/step**

**NOTES !!!!**
~~1. **Training step time  is increased by ~270ms for static shape training**  (576x1024x24) with vae cache (from **7.81s/step to 8.08s/step**). It's probably caused by the changes from `ops` to `mint` operations~~
1. Previous performance on static shape training is intact now.  **mint.repeat_interleave  is 220ms slower than ops.repeat_interleave.  `int(round(S**Tensor(0.5)))` adaptation incurs 50ms degradation.**  
2. To further improve training performance on bucket config, mint.repeat_interleave for mask tiling can be replaced with mint.tile if the target dimension to repeat is 1. **This will be done with MS2.4**
3. jit_level='O1'  is still under fixing for dynamic shape training. **This will be done with MS2.4**

Evaluation:
- [x] static shape training results (512x512x24) on mixkit are visually normal.
- [x] dynamic shape training results (stage 2) , results look fine

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you build and run the code without any errors?

## Who can review?

@xxx


examples/opensora_hpcai/opensora/datasets/video_dataset_refactored.py
```diff
@@ -83,12 +83,12 @@ def __init__(
 
         self.output_columns = output_columns
         if self._buckets is not None:
-            assert vae_latent_folder is None, "`vae_latent_folder` is not supported with bucketing"
+            # assert vae_latent_folder is None, "`vae_latent_folder` is not supported with bucketing"
```
CaitinZhao, line 86: dataset里的assert可以不删吧

SamitHuang: en, fixed

examples/opensora_hpcai/opensora/models/layers/blocks.py
```diff
@@ -483,18 +485,33 @@ class PatchEmbed(nn.Cell):
...
         )
+        self.manual_pad = manual_pad
 
     def construct(self, x: Tensor) -> Tensor:
         b, c, h, w = x.shape
+        if self.manual_pad:
+            # work with pad_mode = valid
+            if h % self.patch_size[0] != 0:
+                pad_h = ops.zeros((b, c, self.patch_size[0] - h % self.patch_size[0], w), x.dtype)
+                x = ops.cat([x, pad_h], 2)
```
hadipash, line 511: Can we just leave padding to be always ON for simplicity? I don't think that it will be a big performance hit.

examples/opensora_hpcai/opensora/models/stdit/stdit2.py
```diff
@@ -385,7 +389,7 @@ def construct(
         T, H, W = self.get_dynamic_size(x)
         S = H * W
         scale = rs / self.input_sq_size
-        base_size = round(S**0.5)
+        base_size = int(round(S ** Tensor(0.5)))
```
hadipash, line 392: `int(Tensor)`?

examples/opensora_hpcai/opensora/schedulers/iddpm/gaussian_diffusion.py
```diff
@@ -299,7 +295,7 @@ def p_sample(
         """
         if frames_mask is not None:
             if frames_mask.shape[0] != x.shape[0]:
-                frames_mask = frames_mask.reshape(1, -1).repeat(2, axis=0)  # HACK
+                frames_mask = repeat_interleave(frames_mask.reshape(1, -1), 2, 0)  # HACK
```
hadipash, line 298: Leave it as it is since `frames_mask` is a numpy array.

examples/opensora_hpcai/scripts/train.py
```diff
@@ -520,6 +525,44 @@ def main(args):
...
+                None,
+            ],
+            dtype=ms.float32,
+        )
+        ar = ms.Tensor(
+            shape=[
+                None,
+            ],
+            dtype=ms.float32,
+        )
```
hadipash, line 562: To turn off ugly `black`'s formatting:
```suggestion
        # fmt: off
        num_frames = ms.Tensor(shape=[None, ], dtype=ms.float32)
        height = ms.Tensor(shape=[None, ], dtype=ms.float32)
        width = ms.Tensor(shape=[None, ], dtype=ms.float32)
        fps = ms.Tensor(shape=[None, ], dtype=ms.float32)
        ar = ms.Tensor(shape=[None, ], dtype=ms.float32)
        # fmt: on
```

zhtmike: can just ust `ms.Tensor(shape=[None], dtype=ms.float32)` with add extra `# fmt: on`

examples/opensora_hpcai/opensora/models/stdit/stdit2.py
zhtmike: > `int(Tensor)`?

int(round(x)) and int() does not behave same

examples/opensora_hpcai/opensora/utils/model_utils.py
```diff
@@ -4,9 +4,20 @@
...
+WHITELIST_OPS = [
+    LayerNorm,
+    Attention,
+    LlamaRMSNorm,
+    SiLU,
+    GELU,
+    GroupNorm,
+    PositionEmbedding2D,
+    SinusoidalEmbedding,
+    T5LayerNorm,
```
hadipash, line 19: `WHITELIST_OPS` are not applied to T5 embedder in `inference.py`.

examples/opensora_hpcai/opensora/models/stdit/stdit2.py
hadipash: 1. `round` already returns `int`. 2. Why need to convert tensor back to int?

examples/opensora_hpcai/scripts/train.py
hadipash: True. I think `[None, ]` is just used to indicate that the array is 2D and the batch dimension is variable.

zhtmike: no it is 1d

examples/opensora_hpcai/opensora/models/stdit/stdit2.py
zhtmike: it is a bit trick,
In graph mode, `round` will be interpreted as `ops.Round` first and try to read a tensor. If it fails (like reading a int) , then it fall backs to the python interpreter silently, and run `round(x)` once and make it as a constant. So to avoid fallback, the input of round should be a Tensor first, that is why added a Tensor(0.5) first instead of 0.5, and the output of round is then a float tensor.

Then, `int` will be interpreted as `ops.Cast`, and converted the float tensor to int tensor. So finally `base_size` is an integer tensor.

P.S., to know what is happening in side the graph, may need to read the IR then ...

hadipash: I see. Well, `base_size` doesn't need to be an `int`

zhtmike: yeah then it should be safe to remove cast

examples/opensora_hpcai/opensora/models/layers/blocks.py
SamitHuang: can not pad empty tensor

examples/opensora_hpcai/opensora/schedulers/iddpm/gaussian_diffusion.py
SamitHuang: fixed

examples/opensora_hpcai/opensora/utils/model_utils.py
SamitHuang: added

examples/opensora_hpcai/scripts/inference.py
```diff
@@ -270,10 +270,12 @@ def main(args):
         )
         text_tokens, mask = Tensor(text_tokens, dtype=ms.int32), Tensor(mask, dtype=ms.uint8)
         text_emb = None
-        if args.dtype in ["fp16", "bf16"]:
-            text_encoder = auto_mixed_precision(text_encoder, amp_level="O2", dtype=dtype_map[args.dtype])
+        # TODO: use FA in T5
+        if args.t5_dtype in ["fp16", "bf16"]:
```
CaitinZhao, line 274: Is this to avoid the accuracy issue in pynative mode？

examples/opensora_hpcai/opensora/utils/model_utils.py
CaitinZhao, line 19: Is this to avoid the accuracy issue in pynative mode？

CaitinZhao (approved):

zhanghuiyao (approved):

vigo999 (approved):

examples/opensora_hpcai/opensora/schedulers/iddpm/gaussian_diffusion.py
hadipash, line 1: `gaussian_diffusion.py` is executed in Pynative mode, I think can keep `assert`s here.

PR #661, opened 2024-09-12
[OpenSora-hpcai] add bucket grouping inside data loader

hadipash:


examples/opensora_hpcai/opensora/datasets/video_dataset_refactored.py
```diff
@@ -289,19 +258,8 @@ def _get_item(self, idx: int) -> Tuple[Any, ...]:
             elif self.video_backend == "cv2":
                 with VideoReader_CV2(video_path) as reader:
```
SamitHuang, line 259: can use `multiprocessing.Process` to allow reading and processing the batch of videos faster?

SamitHuang: generated reference code

```python
def process_video(video_path, output_path, crop_coords):
    # Open the video file
    cap = cv2.VideoCapture(video_path)
    
    # Get the frame count
    frame_count = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
    fps = cap.get(cv2.CAP_PROP_FPS)
    
    # Define the codec and create VideoWriter object
    fourcc = cv2.VideoWriter_fourcc(*'XVID')
    out = cv2.VideoWriter(output_path, fourcc, fps, (crop_coords[2] - crop_coords[0], crop_coords[3] - crop_coords[1]))

    current_frame = 0
    while cap.isOpened():
        ret, frame = cap.read()
        if not ret:
            break
        
        # Crop the frame
        cropped_frame = frame[crop_coords[1]:crop_coords[3], crop_coords[0]:crop_coords[2]]
        
        # Write the cropped frame to the output video
        out.write(cropped_frame)
        
        current_frame += 1
        print(f'Processing frame {current_frame}/{frame_count} from {video_path}')

    # Release everything if job is finished
    cap.release()
    out.release()
    print(f'Finished processing {video_path}')

def main(video_files, crop_coords):
    # Create output directory if it doesn't exist
    output_dir = 'output_videos'
    os.makedirs(output_dir, exist_ok=True)

    # Prepare the arguments for each video processing
    jobs = []
    for video_file in video_files:
        output_file = os.path.join(output_dir, f'cropped_{os.path.basename(video_file)}')
        p = multiprocessing.Process(target=process_video, args=(video_file, output_file, crop_coords))
        jobs.append(p)
        p.start()

    # Wait for all processes to finish
    for job in jobs:
        job.join()
```

hadipash: Good suggestion, I added multithreading to the batch reading (multiprocessing is not trivial and will require a lot of changes):
```python
def get_bucket(self, thw: Tuple[int, int, int], sample_ids: List[int], num_workers: int = 10) -> Tuple[Any, ...]:
    with ThreadPoolExecutor(max_workers=num_workers) as executor:
        batch = list(executor.map(lambda x: self._get_item(x[0], x[1]), zip(sample_ids, [thw] * len(sample_ids))))
    return tuple(np.stack(item) for item in map(list, zip(*batch)))
```
And the reading speed improved a lot:
720p x 51 frames x batch of 3: 36% speedup
240p x 102 frames x batch of 10: 232% speedup.

examples/opensora_hpcai/scripts/train.py
```diff
@@ -529,18 +543,12 @@ def main(args):
...
-    # FIXME: get_dataset_size() is extremely slow when used with bucket_batch_by_length
     if args.bucket_config is None:
         dataset_size = dataloader.get_dataset_size()
     else:
-        # steps per epoch is not constant in bucket config training
-        # FIXME: It is a highly relaxed estimation to ensure enough steps per epoch to sustain training. \
-        # A more precise estimation or run-time infer is to be implemented.
-        dataset_size = math.ceil(num_src_samples / device_num)
+        # Prevent `GeneratorDataset()` from iterating over the whole dataset.
+        dataset_size = 1  # doesn't affect anything when bucketing is enabled
```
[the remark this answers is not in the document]
hadipash: It's indeed confusing, but there's nothing we can do about it since `model.train()` will show an incorrect number of steps per epoch anyway. I find fixing it to 1 to be less confusing than the previous behavior.

PR #701, opened 2024-10-17
[OpenSora-PKU]: add missing files to v0.2.0

wtomin:
Some files were missing from my last PR #698 to synchronize v0.2.0 with the master branch on opensora-pku v1.1.0. Sorry for the trouble.


SamitHuang (approved):
for lpips psnr eval, it would save steps to run them in one script.

wtomin:
> for lpips psnr eval, it would save steps to run them in one script.

I agree with you, but leaving them as two scripts is to align with the code structure with torch v1.1.0.

PR #811, opened 2025-01-03
fix unnessary flake8 E231 check whitespace

HaFred:
The flake8 by default checks whether there are spaces after symbols (in [E231](https://www.flake8rules.com/rules/E231.html)), which may be wrong from time to time for the cases below.
* f-string
* `https://`

This pr fixes it, and also with some refactors for conversion tests. Without this PR, some existing code will be identified as incorrect:
* `examples/latte/sample.py`
* `examples/sv3d/train.py`

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,7 +3,7 @@ max-line-length = 160
 extend-exclude =
     docs/*,
     setup.py
-extend-ignore = E203,E731
+extend-ignore = E203,E731,E231
 per-file-ignores =
     */__init__.py:F401,F403
     examples/*,tests/*:E402
--- a/examples/instantmesh/data/objaverse.py
+++ b/examples/instantmesh/data/objaverse.py
@@ -72,6 +72,7 @@ class ObjaverseDataset:
         total_view_n=32,
         fov=50,
         camera_rotation=True,
+        camera_scaling=False,
     ):
         self.root_dir = Path(root_dir)
         self.input_image_dir = input_image_dir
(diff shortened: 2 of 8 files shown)


HaFred:
Update: This effect turns out to be true for windows python env, not for other unix env like linux/mac os. 
image

There are some other files being affected, including but not limited to：
* https://github.com/mindspore-lab/mindone/blob/master/examples/opensora_hpcai/scripts/train.py#L822
* https://github.com/mindspore-lab/mindone/blob/master/examples/instantmesh/utils/loss_util.py#L48

examples/sv3d/simple_video_sample.py
```diff
@@ -276,11 +278,11 @@ def sample(
         help="path to the input img, or the input imgs dir path",
     )
     parser.add_argument(
-        "--mode", default="1", type=str, help="MindSpore execution mode: Graph mode[0] or Pynative mode[1]"
-    )
-    parser.add_argument(
-        "--device_id", default="2", type=str, help="MindSpore execution mode: Graph mode[0] or Pynative mode[1]"
+        "--mode", default=1, type=int, help="MindSpore execution mode: Graph mode[0] or Pynative mode[1]"
     )
+    parser.add_argument("--device_id", default=2, type=int, help="MindSpore HW device id")
```
SamitHuang, line 283: It's quite weird that the default device is 2. this may conflict the environment variable. better set device id by  `export RANK_ID=2`.

HaFred: How about just removing this argument? This script is expect to run with single card, and I think it's also reasonable that users pass their own `export DEVICE_ID=xx` if they wish to choose a particular NPU card to deploy.

CaitinZhao (approved):

vigo999 (approved):

PR #934, opened 2025-03-26
[DiT]: train script and model updates

wtomin:
1.Some minor modications to make the training script more controlable;
2. change the ops to mint in `mindone/models/dit.py` to make it align with the DiT tutorial.
3. put `train_dit.py` in the working directory which is used in the DiT tutorial.
DiT tutorial to be published: https://github.com/wtomin/tutorials/blob/dit-tutorial/model-examples/dit/code_conversion.md


SamitHuang (approved):

examples/dit/train_dit.py
```diff
@@ -0,0 +1,411 @@
+"""
+DiT training pipeline
+- Image finetuning conditioned on class labels
```
SamitHuang, line 3: better add comments here to explain what is the main difference from train.py and its usage

PR #1022, opened 2025-05-22
[FEATURE] add Cohere model

wtomin:
1. Support Pynative mode inference script of Cohere;
2. Support Pynative + JIT inference script of Cohere;
3. Passed three UTs in pynative mode of Cohere.


SamitHuang (approved):

wtomin:
Cohere will not be included in mindone/transformer. I will close this PR.

PR #1156, opened 2025-07-11
fix diffusers peft lora layer for static graph mode support hotswap

rainbow520lxr:
Questionr:
In MindOne, the static graph hotplug issue of the diffusers library primarily involves the dynamic switching of Lora models. Due to the fact that the adapter_name is fixed as a constant during the static graph compilation, it leads to the failure of subsequent graph computation switching

Solution: In the LoraLayer, an intermediate computation lora_default is added, and during each set_adapter call, the parameters of the specified Lora are copied to lora_default. In the construct method, the computation is entirely delegated to lora_default to produce the final result. It is necessary to ensure that all Lora layers have consistent rank to maintain shape consistency, thereby avoiding recompilation and enabling hot-swapping of parameters.

Performance comparison:
MindOne takes 30 seconds to process a single image in dynamic graph mode and 8 seconds in static graph mode .

### DIFF
--- a/mindone/diffusers/_peft/tuners/lora/layer.py
+++ b/mindone/diffusers/_peft/tuners/lora/layer.py
@@ -27,6 +27,15 @@ from ..tuners_utils import BaseTunerLayer, check_adapters_to_merge
 from .config import LoraConfig
 
 
+def copy_parameters(source_cell, target_cell):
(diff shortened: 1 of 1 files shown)


rainbow520lxr:
好的，期待你们的新版本

